PR #220, opened 2021-09-03
More reliably text if a material needs to be alpha blended

CTOStone:
Relying on TexturedAppearance.isDiffuseColorTextureAlphaBlended is not sufficient to decide when to activate alpha blending.

### DIFF
--- a/core/model-loading/src/main/java/org/lgna/story/resourceutilities/JointedModelGltfExporter.java
+++ b/core/model-loading/src/main/java/org/lgna/story/resourceutilities/JointedModelGltfExporter.java
@@ -43,6 +43,7 @@
 package org.lgna.story.resourceutilities;
 
 import de.javagl.jgltf.impl.v2.*;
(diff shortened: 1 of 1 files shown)


core/model-loading/src/main/java/org/lgna/story/resourceutilities/JointedModelGltfExporter.java
```diff
@@ -73,6 +74,8 @@
 import org.lgna.project.io.JointedModelExporter;
 
 import javax.imageio.ImageIO;
+import java.awt.*;
```
dabeshouse, line 77: Narrow to just `import java.awt.Transparency;`
Importing all of awt makes List ambiguous with java.util.List.
Not sure why this is a problem when I build, but not for you.

dabeshouse (dismissed):
My transparency test worlds look nicer using this code for export.

PR #238, opened 2022-02-17
T1023

weijchen:


pom.xml
```diff
@@ -118,7 +118,7 @@
 
     <module>alice-ide</module>
     <module>netbeans</module>
-    <module>installer</module>
+<!--    <module>installer</module>-->
```
dabeshouse: Changes to pom.xml should not be part of this PR.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,184 +55,148 @@
...
-    //c.gl.glColor3d( 1,1,1 );
     int id = 0;
+    int numberOfCorner = 4;
+    float[][] leftTexCoord = {{.75f, .66f}, {.75f, .33f}, {.5f, .33f}, {.5f, .66f}};
+    float[][] rightTexCoord = {{.25f, .66f}, {.25f, .33f}, {.0f, .33f}, {.0f, .66f}};
+    float[][] bottomTexCoord = {{.5f, .66f}, {.25f, .66f}, {.25f, 1.0f}, {.5f, 1.0f}};
+    float[][] topTexCoord = {{.5f, .0f}, {.25f, .0f}, {.25f, .33f}, {.5f, .33f}};
+    float[][] frontTexCoord = {{.5f, .33f}, {.25f, .33f}, {.25f, .66f}, {.5f, .66f}};
+    float[][] backTexCoord = {{.75f, .66f}, {1.0f, .66f}, {1.0f, .33f}, {.75f, .33f}};
+    double[][] leftVertex = {{this.xMin, this.yMin, this.zMax}, {this.xMin, this.yMax, this.zMax}, {this.xMin, this.yMax, this.zMin}, {this.xMin, this.yMin, this.zMin}};
```
dabeshouse: Consider renaming to leftFace, since these are the faces of the box, built out of four vertices.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,184 +55,148 @@
...
+    vertexMap.put("bottom", bottomVertex);
+    vertexMap.put("top", topVertex);
+    vertexMap.put("front", frontVertex);
+    vertexMap.put("back", backVertex);
 
     c.gl.glBegin(GL_QUADS);
+
+    // xMin face
+    // c.gl.glColor3d( 1,1,1 );
+    float[][] leftSideTex = texCoordMap.get("left");
```
dabeshouse: The next twelve lines are the same for all eight faces, except for the variable names and the normal for lighting.
You've gone this far, you could extract a method and eliminate a lot of repetition.

dabeshouse (changes requested):
Restoring the pom is essential. My other suggestions are just that, suggestions.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,186 +55,99 @@
...
-    c.gl.glVertex3d(this.xMin, this.yMin, this.zMin);
+  private void DrawBox(Context c, int id, boolean isSubElementRequired, boolean isLightingEnabled, boolean isTexturingEnabled, int numberOfCorner, String side, Map<String, int[]> normalArr, Map<String, float[][]> texMat, Map<String, double[][]> vertexMat) {
+    int[] sideNormal3d = normalArr.get(side);
+    float[][] sideTex = texMat.get(side);
+    double[][] sideVertex = vertexMat.get(side);
 
-    //zMax face
-    //c.gl.glColor3d( 0,0,1 );
     if (isSubElementRequired) {
       c.gl.glLoadName(id++);
```
dabeshouse: With this moved into a method the increment only happens locally, so if subElement is required this will result in the same id on each. Move the increment out to glBox and the ids will be distinct.

(That said, I looked for a code path where isSubElementRequired is true and can not find one. There may be a way under a certain recovery path for picking having failed, but my cursory hunt didn't find it. Cleaning that up is task for another day.)

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,186 +55,99 @@
...
- * Texture is defined by glTexCoord2f, which uses two value to crop the original picture
+ * Texture is defined by glTexCoord2f, which uses two values to crop the original picture
  */
 public class GlrBox extends GlrShape<Box> {
   private void glBox(Context c, boolean isLightingEnabled, boolean isSubElementRequired, boolean isTexturingEnabled) {
     //todo: revist vertex ordering
-    //xMin face
-    //c.gl.glColor3d( 1,1,1 );
     int id = 0;
+    int numberOfCorner = 4;
```
dabeshouse: These values are constant and don't need to be redeclared with every call. If you move them out of the method onto the class they also won't have to be passed into each drawBox call.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,186 +55,99 @@
...
-    }
-    c.gl.glVertex3d(this.xMax, this.yMin, this.zMin);
+    c.gl.glEnd();
+  }
 
-    if (isTexturingEnabled) {
-      c.gl.glTexCoord2f(1.0f, .66f);
-    }
-    c.gl.glVertex3d(this.xMin, this.yMin, this.zMin);
+  private void DrawBox(Context c, int id, boolean isSubElementRequired, boolean isLightingEnabled, boolean isTexturingEnabled, int numberOfCorner, String side, Map<String, int[]> normalArr, Map<String, float[][]> texMat, Map<String, double[][]> vertexMat) {
```
dabeshouse: Rename to drawBox, to go with Java code style over C#

dabeshouse (changes requested):

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,266 +55,195 @@
...
+        //todo: revist vertex ordering
+        int id = 0;
+        double[][] leftFace = {{this.xMin, this.yMin, this.zMax}, {this.xMin, this.yMax, this.zMax}, {this.xMin, this.yMax, this.zMin}, {this.xMin, this.yMin, this.zMin}};
+        double[][] rightFace = {{this.xMax, this.yMin, this.zMin}, {this.xMax, this.yMax, this.zMin}, {this.xMax, this.yMax, this.zMax}, {this.xMax, this.yMin, this.zMax}};
+        double[][] bottomFace = {{this.xMin, this.yMin, this.zMin}, {this.xMax, this.yMin, this.zMin}, {this.xMax, this.yMin, this.zMax}, {this.xMin, this.yMin, this.zMax}};
+        double[][] topFace = {{this.xMin, this.yMax, this.zMax}, {this.xMax, this.yMax, this.zMax}, {this.xMax, this.yMax, this.zMin}, {this.xMin, this.yMax, this.zMin}};
+        double[][] frontFace = {{this.xMin, this.yMax, this.zMin}, {this.xMax, this.yMax, this.zMin}, {this.xMax, this.yMin, this.zMin}, {this.xMin, this.yMin, this.zMin}};
+        double[][] backFace = {{this.xMin, this.yMin, this.zMax}, {this.xMax, this.yMin, this.zMax}, {this.xMax, this.yMax, this.zMax}, {this.xMin, this.yMax, this.zMax}};
+        normal3dMap.put("left", leftNormal3d);
```
dabeshouse: Filling in the first normal and tex maps can be done once in a static initializer where they are declared below:
```
  …
  private static final Map<String, int[]> normal3dMap = new HashMap<>();
  private static final Map<String, float[][]> texCoordMap = new HashMap<>();
  private static final Map<String, double[][]> vertexMap = new HashMap<>();
  static {
    normal3dMap.put("left", leftNormal3d);
    …
  }
```
Only the vertexMap values are dynamic.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,266 +55,195 @@
...
+        vertexMap.put("top", topFace);
+        vertexMap.put("front", frontFace);
+        vertexMap.put("back", backFace);
+
+        c.gl.glBegin(GL_QUADS);
+
+        // xMin face
+        // c.gl.glColor3d( 1,1,1 );
+        if (isSubElementRequired) {
+            c.gl.glLoadName(id++);
```
dabeshouse: In my previous comment about incrementing this I had intended it to become part of the method call:
`drawBox(c, id++…`
But this works.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,266 +55,195 @@
...
+    private static final int[] topNormal3d = {0, 1, 0};
+    private static final int[] frontNormal3d = {0, 0, -1};
+    private static final int[] backNormal3d = {0, 0, 1};
+    private static final float[][] leftTexCoord = {{.75f, .66f}, {.75f, .33f}, {.5f, .33f}, {.5f, .66f}};
+    private static final float[][] rightTexCoord = {{.25f, .66f}, {.25f, .33f}, {.0f, .33f}, {.0f, .66f}};
+    private static final float[][] bottomTexCoord = {{.5f, .66f}, {.25f, .66f}, {.25f, 1.0f}, {.5f, 1.0f}};
+    private static final float[][] topTexCoord = {{.5f, .0f}, {.25f, .0f}, {.25f, .33f}, {.5f, .33f}};
+    private static final float[][] frontTexCoord = {{.5f, .33f}, {.25f, .33f}, {.25f, .66f}, {.5f, .66f}};
+    private static final float[][] backTexCoord = {{.75f, .66f}, {1.0f, .66f}, {1.0f, .33f}, {.75f, .33f}};
+    private static Map<String, int[]> normal3dMap = new HashMap<>();
```
dabeshouse: The maps can be made `final` as well.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrBox.java
```diff
@@ -55,266 +55,195 @@
...
+    private static Map<String, float[][]> texCoordMap = new HashMap<>();
+    private static Map<String, double[][]> vertexMap = new HashMap<>();
```
dabeshouse: This being `static` means the same map is used for all boxes in the scene.
It probably should not be.

In theory it could be possible for multiple cubes to overwrite each and interfere with rendering.

dabeshouse (changes requested):
Please restore this to two space indents instead of four. A meaningless arbitrary thing, but it makes comparisons easier. Also, please change your editor settings to use two spaces as well.

PR #241, opened 2022-03-07
T1019 [Shapes-IDE] Imported image painted on a sphere in Alice is flipped and tilted

weijchen:


core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/Context.java
```diff
@@ -83,7 +83,6 @@ public void incrementScaledCount() {
     if (this.scaledCount == 1) {
       this.enableNormalize();
     }
-
```
dabeshouse, line 86: Try to avoid including files with this minor a change.

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrSphere.java
```diff
@@ -61,22 +64,118 @@
...
+      }
+      gl.glEnd();
+    }
   }
 
   @Override
   protected void renderGeometry(RenderContext rc, GlrVisual.RenderType renderType) {
-    //Required for quadric shapes like spheres, discs, and cylinders
     boolean isTextureEnabled = rc.isTextureEnabled();
     rc.glu.gluQuadricTexture(rc.getQuadric(), isTextureEnabled);
```
dabeshouse, line 172: I think this can be removed, since the Quadric is not being used

PR #242, opened 2022-03-07
Feature/cylinder texture

dabeshouse:


core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrDisc.java
```diff
@@ -100,7 +138,7 @@ protected void pickGeometry(PickContext pc, boolean isSubElementRequired) {
       name = -1;
     }
     pc.gl.glPushName(name);
-    glDisc(pc);
+    glDisc(pc, false);
```
weijchen, line 141: Is this supposed to be `glDisc(pc, true);` ?

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrCylinder.java
```diff
@@ -135,26 +186,31 @@ protected void pickGeometry(PickContext pc, boolean isSubElementRequired) {
       name = -1;
     }
     pc.gl.glPushName(name);
-    glCylinder(pc);
+    glCylinder(pc, false);
```
weijchen, line 186: also, the second argument should be true?

dabeshouse: This in `pickGeometry` that is used to determine interactions with the scene, so the full rendering is not required. In `renderGeometry` the isTextureEnabled does get checked and used (and is probably no longer needed above on line 125)

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrDisc.java
dabeshouse: This in pickGeometry that is used to determine interactions with the scene, so the full rendering is not required. In renderGeometry the isTextureEnabled does get checked and used (and is probably no longer needed above on line 125)

core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrCylinder.java
dabeshouse: Again, pickGeometry not renderGeometry, but this time I did remove the isTextureEnabled check in renderGeometry.

PR #261, opened 2022-06-19
remove assert

weijchen:
remove assert line

### DIFF
--- a/core/ide/src/main/java/org/alice/stageide/sceneeditor/interact/croquet/AbstractSetLocalTransformationActionOperation.java
+++ b/core/ide/src/main/java/org/alice/stageide/sceneeditor/interact/croquet/AbstractSetLocalTransformationActionOperation.java
@@ -77,6 +77,7 @@ public abstract class AbstractSetLocalTransformationActionOperation extends Abst
 
   protected void setLocalTransformation(AffineMatrix4x4 lt) {
     AbstractTransformable sgTransformable = getSGTransformable();
+    if (sgTransformable == null) return;
     if (this.getAnimator() != null) {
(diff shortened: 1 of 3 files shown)


core/story-api/src/main/java/edu/cmu/cs/dennisc/animation/affine/AffineAnimation.java
```diff
@@ -66,7 +66,6 @@ public AbstractTransformable getSubject() {
   }
 
   public void setSubject(AbstractTransformable sgSubject) {
-    assert sgSubject != null;
```
dabeshouse, line 69: If the value is null it triggers a different red queen that should be avoided as well.
More elaboration on https://phab.alice.org/T1090

dabeshouse (changes requested):
Having a different red queen is no better for the user and makes tracking the issue more difficult.
Can changes be made to prevent further error pop ups?

core/ide/src/main/java/org/alice/stageide/sceneeditor/interact/croquet/AbstractSetLocalTransformationActionOperation.java
```diff
@@ -77,6 +77,7 @@ public AbstractSetLocalTransformationActionOperation(Group group, UUID individua
 
   protected void setLocalTransformation(AffineMatrix4x4 lt) {
     AbstractTransformable sgTransformable = getSGTransformable();
+    if (sgTransformable == null) return;
```
dabeshouse: A tiny thing, but please add brackets around the return since this violates the style check and prevents the build.

PR #262, opened 2022-07-06
unify usage of integer and floating number

weijchen:


core/ast/src/main/resources/org/lgna/project/ast/ArithmeticInfixExpression.properties
```diff
@@ -1,7 +1,7 @@
-INTEGER_DIVIDE = </leftOperand/> INTEGER_DIVIDE </rightOperand/>
-INTEGER_REMAINDER = </leftOperand/> INTEGER_REMAINDER </rightOperand/>
+INTEGER_DIVIDE = </leftOperand/> WHOLENUMBER_DIVIDE </rightOperand/>
```
dabeshouse: Please add an _ in front of NUMBER to make make it WHOLE_NUMBER_DIVIDE (and for others below)
It is more readable and consistent with the camel caps WholeNumber.
(I am not keen on either the ALL CAPS, or the length, but can't offer anything better, so adding _s is enough)

dabeshouse (changes requested):
All properties files changed, and almost done.
One more small tweak before merge.

core/ast/src/main/resources/org/lgna/project/ast/ArithmeticInfixExpression.properties
dabeshouse: I made the changes

PR #270, opened 2022-07-22
T1207

weijchen:


core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
+    topTransform.orientation.right.set(-1, 0, 0);
+    topTransform.orientation.backward.set(0, 1, 0);
+    this.topOrthoMarkerImp.setLocalTransformation(topTransform);
+    picturePlane.setCenter(0, 0);
+    picturePlane.setHeight(16);
+    this.topOrthoMarkerImp.setPicturePlane(picturePlane);
+
+    AffineMatrix4x4 sideTransform = AffineMatrix4x4.createIdentity();
+    sideTransform.translation.x = targetTransform.translation.x + 10.0f;
+    sideTransform.translation.y = targetHeight / 2;
```
dabeshouse: This makes sense when the target is on the ground, but breaks when placed higher or lower. Try translation.y and look at shapes with 0 at the base (bipeds) and inside the model (fish)

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
+    sideTransform.translation.z = targetTransform.translation.z;
+    sideTransform.orientation.setValue(new ForwardAndUpGuide(Vector3.accessNegativeXAxis(), Vector3.accessPositiveYAxis()));
+    this.sideOrthoMarkerImp.setLocalTransformation(sideTransform);
+    picturePlane.setHeight(6);
+    this.sideOrthoMarkerImp.setPicturePlane(picturePlane);
+
+    AffineMatrix4x4 frontTransform = AffineMatrix4x4.createIdentity();
+    frontTransform.translation.x = targetTransform.translation.x;
+    frontTransform.translation.y = targetHeight / 2;
```
dabeshouse: Same as side view - use translation

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
+
+    AffineMatrix4x4 topTransform = AffineMatrix4x4.createIdentity();
+    topTransform.translation.x = targetTransform.translation.x;
+    topTransform.translation.y = 10;
+    topTransform.translation.z = targetTransform.translation.z;
+    topTransform.orientation.up.set(0, 0, 1);
+    topTransform.orientation.right.set(-1, 0, 0);
+    topTransform.orientation.backward.set(0, 1, 0);
+    this.topOrthoMarkerImp.setLocalTransformation(topTransform);
+    picturePlane.setCenter(0, 0);
```
dabeshouse, line 482: Since this picture place is the same as is initially set I don't think you need to repeat these three lines.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
+    picturePlane.setHeight(16);
+    this.topOrthoMarkerImp.setPicturePlane(picturePlane);
+
+    AffineMatrix4x4 sideTransform = AffineMatrix4x4.createIdentity();
+    sideTransform.translation.x = targetTransform.translation.x + 10.0f;
+    sideTransform.translation.y = targetHeight / 2;
+    sideTransform.translation.z = targetTransform.translation.z;
+    sideTransform.orientation.setValue(new ForwardAndUpGuide(Vector3.accessNegativeXAxis(), Vector3.accessPositiveYAxis()));
+    this.sideOrthoMarkerImp.setLocalTransformation(sideTransform);
+    picturePlane.setHeight(6);
```
dabeshouse: For side and front you have overridden the original picture plane height.
If you change the values in the initialization you can drop these as repeats and have no picturePlane logic in this method.

core/ide/src/main/java/org/alice/stageide/sceneeditor/viewmanager/CameraMarkerTracker.java
```diff
@@ -164,18 +164,18 @@ public void startTrackingCameraView(View view) {
...
-  private void setCameraToSelectedMarker() {
-    stopTrackingCamera();
+  public void setCameraToSelectedMarker() {
     if (this.activeMarker instanceof OrthographicCameraMarkerImp) {
       OrthographicCameraMarkerImp orthoMarker = (OrthographicCameraMarkerImp) this.activeMarker;
       sceneEditor.switchToOrthographicCamera();
       Transformable cameraParent = (Transformable) CameraMarkerTracker.this.orthographicCamera.getParent();
       CameraMarkerTracker.this.orthographicCamera.picturePlane.setValue(new ClippedZPlane(orthoMarker.getPicturePlane()));
-      cameraParent.setTransformation(CameraMarkerTracker.this.activeMarker.getTransformation(org.lgna.story.implementation.AsSeenBy.SCENE), CameraMarkerTracker.this.orthographicCamera.getRoot());
+      cameraParent.setTransformation(CameraMarkerTracker.this.activeMarker.getTransformation(org.lgna.story.implementation.AsSeenBy.SELF), CameraMarkerTracker.this.orthographicCamera.getRoot());
```
dabeshouse, line 178: Good catch here on the reference frame

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
   }
 
+  private void updateOrthographicCameraView(EntityImp targetImplementation) {
+    double targetHeight = targetImplementation.getAxisAlignedMinimumBoundingBox().getHeight();
+    AffineMatrix4x4 targetTransform = targetImplementation.getAbsoluteTransformation();
+    ClippedZPlane picturePlane = new ClippedZPlane();
+
+    AffineMatrix4x4 topTransform = AffineMatrix4x4.createIdentity();
+    topTransform.translation.x = targetTransform.translation.x;
+    topTransform.translation.y = 10;
```
dabeshouse: Instead of a constant 10 this could be based on the model height to better handle very large or small models. Maybe 3x height?
If you wanted to get really fancy you could  make sure the model takes up some portion of the view, but that might get complicated.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
+    topTransform.orientation.up.set(0, 0, 1);
+    topTransform.orientation.right.set(-1, 0, 0);
+    topTransform.orientation.backward.set(0, 1, 0);
+    this.topOrthoMarkerImp.setLocalTransformation(topTransform);
+    picturePlane.setCenter(0, 0);
+    picturePlane.setHeight(16);
+    this.topOrthoMarkerImp.setPicturePlane(picturePlane);
+
+    AffineMatrix4x4 sideTransform = AffineMatrix4x4.createIdentity();
+    sideTransform.translation.x = targetTransform.translation.x + 10.0f;
```
dabeshouse: Again, the constant 10 could instead be based on the width of the axis aligned box built at the beginning of the method.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -463,6 +475,44 @@ public void setSelectedExpression(Expression expression) {
...
+    sideTransform.translation.z = targetTransform.translation.z;
+    sideTransform.orientation.setValue(new ForwardAndUpGuide(Vector3.accessNegativeXAxis(), Vector3.accessPositiveYAxis()));
+    this.sideOrthoMarkerImp.setLocalTransformation(sideTransform);
+    picturePlane.setHeight(6);
+    this.sideOrthoMarkerImp.setPicturePlane(picturePlane);
+
+    AffineMatrix4x4 frontTransform = AffineMatrix4x4.createIdentity();
+    frontTransform.translation.x = targetTransform.translation.x;
+    frontTransform.translation.y = targetHeight / 2;
+    frontTransform.translation.z = targetTransform.translation.z - 10.0f;
```
dabeshouse: Replace 10 with value based on AAB.width

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -453,6 +454,17 @@ private void setSelectedInstance(InstanceFactory instanceFactory) {
     getPropertyPanel().setSelectedInstance(instanceFactory);
   }
 
+  private void triggerOrthographicCameraChange(Expression selectedFieldExpression) {
```
dabeshouse, line 444: This is called on camera and model selection and that seems to work in general.
If the active model is moved or resized it can end up outside of the view.
Not sure if the additional tracking is worth it, but I observed the effect and want you to be aware of it.

dabeshouse (changes requested):

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: I make it 3x height. Yeah, that is a good suggestion! We can add the portion one in the future haha

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -477,38 +477,33 @@ public void setSelectedExpression(Expression expression) {
 
   private void updateOrthographicCameraView(EntityImp targetImplementation) {
     double targetHeight = targetImplementation.getAxisAlignedMinimumBoundingBox().getHeight();
+    double targetWidth = targetImplementation.getAxisAlignedMinimumBoundingBox().getWidth();
```
dabeshouse: Unfortunately, getAxisAlignedMinimumBoundingBox() can be a little expensive, so I recommend calling it only once and storing it locally.
Additionally, use getDynamicAxisAlignedMinimumBoundingBox(AsSeenBy.SCENE) so the box constructed aligns with the views used here.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -477,38 +477,33 @@ public void setSelectedExpression(Expression expression) {
...
+    double targetDepth = targetImplementation.getAxisAlignedMinimumBoundingBox().getDepth();
+    double defaultDistance = 10.0f;
     AffineMatrix4x4 targetTransform = targetImplementation.getAbsoluteTransformation();
-    ClippedZPlane picturePlane = new ClippedZPlane();
 
     AffineMatrix4x4 topTransform = AffineMatrix4x4.createIdentity();
     topTransform.translation.x = targetTransform.translation.x;
-    topTransform.translation.y = 10;
+    topTransform.translation.y = targetHeight != 0 ? targetHeight * 10 : defaultDistance;
```
dabeshouse: This does not consider where the model is vertically. Add translation.y to see things in the air or under ground.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -477,38 +477,33 @@ public void setSelectedExpression(Expression expression) {
...
     topTransform.translation.z = targetTransform.translation.z;
     topTransform.orientation.up.set(0, 0, 1);
     topTransform.orientation.right.set(-1, 0, 0);
     topTransform.orientation.backward.set(0, 1, 0);
     this.topOrthoMarkerImp.setLocalTransformation(topTransform);
-    picturePlane.setCenter(0, 0);
```
dabeshouse: I steered you wrong in removing these. The in/out arrows affect this, not the position of the camera, so using these arrows lead to persistent effects across views.
All three should setPicturePlane calls be restored and amended.
Instead of using fixed values, set them based on the model, perhaps plane.setHeight(2 * targetHeight)
From above the depth of the object corresponds to the height of the windows, so use that here.

dabeshouse (changes requested):
A few more tweak requests.
I see you brought in commits from develop. If you want to do that I highly recommend rebasing your branch onto develop instead. It keeps the branch activity clearer and prevents duplicate commits in the overall history. If you want you can cherry pick and recreate your branch, or just think about it next time.

dabeshouse (approved):

dabeshouse:
I got fiddly and made a few more changes directly to your branch. I tried to split the commits cleanly so you can see each change I made.
If these are OK by you we can merge this in.

weijchen:
Thank you! I think we can merge it.

PR #271, opened 2022-07-25
T1279 Torus. Add uniform resize. Make others consistent. Update values.

dabeshouse:


weijchen (approved):
lgtm, we can merge the change

PR #276, opened 2022-08-13
add functionality to automatically center layout camera; rename persp…

weijchen:
…ective cams in storytellingSceneEditor to make it easier to read; add cameratype

### DIFF
--- a/core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
+++ b/core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
@@ -582,7 +582,7 @@ public class StorytellingSceneEditor extends AbstractSceneEditor implements Rend
 
     PerspectiveCameraMarker layoutCameraMarker = new PerspectiveCameraMarker();
     layoutCameraMarker.setColorId(Color.LIGHT_BLUE);
-    this.layoutSceneMarkerImp= EmployeesOnly.getImplementation(layoutCameraMarker);
+    this.layoutSceneMarkerImp = EmployeesOnly.getImplementation(layoutCameraMarker);
     this.layoutSceneMarkerImp.setDisplayVisuals(true);
     this.layoutSceneMarkerImp.setCameraType(LAYOUT_CAMERA);
     MarkerUtilities.addIconForCameraImp(this.layoutSceneMarkerImp, "sceneEditorCamera");
--- a/core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
+++ b/core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
@@ -154,7 +154,6 @@ import org.lgna.story.resources.ModelResource;
  *
  */
 public class StorytellingSceneEditor extends AbstractSceneEditor implements RenderTargetListener {
-
   private class SceneEditorDropReceptor extends AbstractDropReceptor {
     @Override
     public boolean isPotentiallyAcceptingOf(DragModel dragModel) {
@@ -342,8 +341,8 @@ public class StorytellingSceneEditor extends AbstractSceneEditor implements Rend
   private OrthographicCameraMarkerImp frontOrthoMarkerImp = null;
   private OrthographicCameraMarkerImp sideOrthoMarkerImp = null;
   private List<OrthographicCameraMarkerImp> orthographicCameraMarkerImps = new LinkedList<OrthographicCameraMarkerImp>();
-  private PerspectiveCameraMarkerImp openingSceneMarkerImp;
-  private PerspectiveCameraMarkerImp sceneViewMarkerImp;
+  private PerspectiveCameraMarkerImp startingCameraMarkerImp;
+  private PerspectiveCameraMarkerImp layoutSceneMarkerImp;
 
   private ComboBox<View> mainCameraViewSelector;
   private CameraMarkerTracker mainCameraViewTracker;
@@ -368,7 +367,7 @@ public class StorytellingSceneEditor extends AbstractSceneEditor implements Rend
   }
 
   public void setStartingCameraMarkerTransformation(AffineMatrix4x4 transform) {
-    openingSceneMarkerImp.setLocalTransformation(transform);
+    startingCameraMarkerImp.setLocalTransformation(transform);
   }
 
   public static class SceneEditorProgramImp extends ProgramImp {
@@ -424,7 +423,7 @@ public class StorytellingSceneEditor extends AbstractSceneEditor implements Rend
       if (field instanceof UserField) {
         UserField uf = (UserField) field;
         setSelectedField(uf.getDeclaringType(), uf);
-        triggerOrthographicCameraChange(expression);
+        triggerCameraChange(expression);
       }
     } else if (expression instanceof MethodInvocation) {
       setSelectedExpression(expression);
@@ -441,14 +440,14 @@ public class StorytellingSceneEditor extends AbstractSceneEditor implements Rend
     getPropertyPanel().setSelectedInstance(instanceFactory);
   }
 
-  private void triggerOrthographicCameraChange(Expression selectedFieldExpression) {
+  private void triggerCameraChange(Expression selectedFieldExpression) {
     Object instanceInJava = IDE.getActiveInstance().getSceneEditor().getInstanceInJavaVMForExpression(selectedFieldExpression);
     if (instanceInJava instanceof SThing) {
-        if ((instanceInJava instanceof SCamera) || (instanceInJava instanceof SGround)) {
-          return;
-        }
-        EntityImp implementation = EmployeesOnly.getImplementation((SThing) instanceInJava);
-        updateOrthographicCameraView(implementation);
+      if ((instanceInJava instanceof SCamera) || (instanceInJava instanceof SGround)) {
+        return;
+      }
+      EntityImp implementation = EmployeesOnly.getImplementation((SThing) instanceInJava);
+      updateCameraView(implementation);
     }
   }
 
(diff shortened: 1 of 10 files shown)


core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -342,19 +348,19 @@ public void valueChanged(ValueEvent<InstanceFactory> e) {
...
   private CameraMarkerTracker mainCameraViewTracker;
   private View savedSceneEditorViewSelection = null;
 
-  private ValueListener<View> mainCameraViewSelectionObserver = new ValueListener<View>() {
-    @Override
-    public void valueChanged(ValueEvent<View> e) {
-      StorytellingSceneEditor.this.handleMainCameraViewSelection(mainCameraViewTracker.getCameraMarker(e.getNextValue()));
-    }
-  };
+  public static int STARTING_CAMERA = 0;
```
dabeshouse, line 358: Instead of numbers an enum can make it clearer.
If you use the existing View enum, and set it on all five MarkerImps it could simplify some other logic.
(I find View to be a less than ideal name, so if you come up with a better one feel free to change it)

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -969,30 +977,21 @@ protected void setActiveScene(UserField sceneField) {
...
-
-      //      SideComposite.getInstance().getCameraMarkersTab().getMarkerListState().addAndInvokeValueListener( this.cameraMarkerFieldSelectionObserver );
-      //      SideComposite.getInstance().getObjectMarkersTab().getMarkerListState().addAndInvokeValueListener( this.objectMarkerFieldSelectionObserver );
-      //      ManagedCameraMarkerFieldState.getInstance( (NamedUserType)sceneAliceInstance.getType() ).addAndInvokeValueListener( this.cameraMarkerFieldSelectionObserver );
-      //      ManagedObjectMarkerFieldState.getInstance( (NamedUserType)sceneAliceInstance.getType() ).addAndInvokeValueListener( this.objectMarkerFieldSelectionObserver );
 
       for (AbstractField field : sceneField.getValueType().getDeclaredFields()) {
-        //Turn markers on so they're visible in the scene editor (note: markers are hidden by default so that when a world runs they aren't scene. we have to manually make them visible to see them in the scene editor)
+        // Turn markers on, so they're visible in the scene editor (note: markers are hidden by default so that when a world runs they aren't scene.
```
dabeshouse, line 993: The spelling error was already there, but should end with "… they aren't seen."

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -555,17 +574,19 @@ private boolean isSelectableType(AbstractType<?, ?, ?> valueType) {
...
-    MarkerUtilities.setViewForCameraImp(this.sceneViewMarkerImp, View.LAYOUT_SCENE_VIEW);
+    this.startingCameraMarkerImp = EmployeesOnly.getImplementation(openingSceneMarker);
+    this.startingCameraMarkerImp.setDisplayVisuals(true);
+    this.startingCameraMarkerImp.setCameraType(STARTING_CAMERA);
+    MarkerUtilities.addIconForCameraImp(this.startingCameraMarkerImp, "mainCamera");
+    MarkerUtilities.setViewForCameraImp(this.startingCameraMarkerImp, View.STARTING_CAMERA_VIEW);
+
+    PerspectiveCameraMarker layoutCameraMarker = new PerspectiveCameraMarker();
+    layoutCameraMarker.setColorId(Color.LIGHT_BLUE);
+    this.layoutSceneMarkerImp= EmployeesOnly.getImplementation(layoutCameraMarker);
```
dabeshouse: Add a space before the equals to make the style check happy

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -555,17 +574,19 @@ private boolean isSelectableType(AbstractType<?, ?, ?> valueType) {
...
-    sceneViewMarker.setColorId(Color.LIGHT_BLUE);
-    this.sceneViewMarkerImp = EmployeesOnly.getImplementation(sceneViewMarker);
-    this.sceneViewMarkerImp.setDisplayVisuals(true);
-    MarkerUtilities.addIconForCameraImp(this.sceneViewMarkerImp, "sceneEditorCamera");
-    MarkerUtilities.setViewForCameraImp(this.sceneViewMarkerImp, View.LAYOUT_SCENE_VIEW);
+    this.startingCameraMarkerImp = EmployeesOnly.getImplementation(openingSceneMarker);
+    this.startingCameraMarkerImp.setDisplayVisuals(true);
+    this.startingCameraMarkerImp.setCameraType(STARTING_CAMERA);
+    MarkerUtilities.addIconForCameraImp(this.startingCameraMarkerImp, "mainCamera");
+    MarkerUtilities.setViewForCameraImp(this.startingCameraMarkerImp, View.STARTING_CAMERA_VIEW);
```
dabeshouse: If you set a View on each MarkerImp the second argument here is not needed. Not much of a simplification, but an option.

dabeshouse (changes requested):
These changes look good and seem to make undo/redo behave better.
Just asking for a tweak for the style checker and a suggestion to use an enum.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: I'm thinking of CameraOption or CameraAngle. And I also want to change STARTING_CAMERA_VIEW to STARTING_CAMERA and LAYOUT_SCENE_VIEW to LAYOUT_CAMERA. Also for the top, side, and front to make them more aligned. What do you think?

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: Why does the second argument here is not needed?

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
dabeshouse: I am good with `CameraOption`. Angle has too many implications. Also, simply adding `Camera` to call it `CameraView` could work.
Dropping VIEW from the two enum values is more consistent.
The enums are the names in the code. The user facing strings are fetched in `MarkerUtilities.getNameForView()` where it looks up the user facing name in `cameraViews.properties`. Not sure if you intend to change those too, but I am not a fan of our case choices there.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
dabeshouse: If you change `setCameraType()` to take a `View` (or whatever it gets called) enum instead of a number, and do that for all the MarkerImps, the markerImp passed in to `setViewForCameraImp` can be queried with `getCameraType()` instead of passing it in.
This is perhaps moe an example of of some possible benefit in using the enum instead of an int.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: Got it, I'll rename it `CameraOption`
I'm not a fan of the choices too. Terms like "Layout Scene" and "Starting Camera" seems not relevant to me. Can I also rename these properties? If yes, I'll drop View on both selections and rename Layout Scene to Layout Camera.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: Got it! we should use some benefits of the enum.

dabeshouse:
As you found out, my suggestion would lead to circular dependencies.
We should not track the IDE CameraOption on the core Markers.
You can roll back your initial solution, with a number to flag the main camera, or you can unravel some more of the Alice code in and around MarkerUtilities.

A `StorytellingSceneEditor` has a `CameraMarkerTracker mainCameraViewTracker;`
That in turn holds a `Map<CameraOption, CameraMarkerImp> mapViewToMarker` that is set up by the `StorytellingSceneEditor`

`MarkerUtilities` is a static grab bag of code and data. It has `HashMap<CameraMarker, CameraOption> cameraToViewMap` going the other way from marker to option. This is what is being built when you call `setViewForCameraImp` and then used at the bottom of the method when you call `getNameForCameraImp`  in lines like
`this.startingCameraMarkerImp.getAbstraction().setName(MarkerUtilities.getNameForCameraImp(this.startingCameraMarkerImp));`
If each of these move up to where they are constructed you can change them to more like:
`this.startingCameraMarkerImp.getAbstraction().setName(MarkerUtilities.getNameForView(CameraOption.STARTING_CAMERA));`

Changing that leaves the only users of `cameraToViewMap` being `MoveMarkerToActiveCameraActionOperation` and `MoveActiveCameraToMarkerActionOperation` which looks up a key to generate a localized tooltip for the camera marker buttons.
I notice that those buttons and tooltips have a few problems. Just wrote up T1298 and T1299.

If you want to keep fixing we can talk more. If you want to roll back to what worked, we can do that.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
dabeshouse: The build runs checkstyle. If this remains `this.layoutSceneMarkerImp=` instead of `this.layoutSceneMarkerImp =` it will not build.
Trivial, but please fix.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
dabeshouse: The rename is good, but my original comment remains. Marker has a redundant type that is an integer supported by these two integers instead of using `CameraOption`.

dabeshouse (changes requested):
Two things left to resolve, thanks

dabeshouse (approved):

PR #286, opened 2022-11-19
add clamp value function, enlarge the zoom scale for all camera views

weijchen:


core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -468,20 +470,27 @@ public void setSelectedExpression(Expression expression) {
...
     AffineMatrix4x4 layoutTransform = AffineMatrix4x4.createIdentity();
+    double CAMERA_CLAMP_MAX = 100.0;
+    double CAMERA_CLAMP_MIN = 2.0;
+    double PICTURE_CLAMP_MAX = 100.0;
+    double PICTURE_CLAMP_MIN = 1.0;
     layoutTransform.translation.x = targetTransform.translation.x;
-    layoutTransform.translation.y = targetTransform.translation.y + DEFAULT_LAYOUT_CAMERA_Y_OFFSET;
-    layoutTransform.translation.z = targetTransform.translation.z - DEFAULT_LAYOUT_CAMERA_Z_OFFSET;
+    layoutTransform.translation.y = targetTransform.translation.y + (targetHeight != 0 ? clampValue(CAMERA_CLAMP_MIN, CAMERA_CLAMP_MAX, targetHeight * DEFAULT_LAYOUT_CAMERA_Y_OFFSET) : DEFAULT_LAYOUT_CAMERA_Y_OFFSET);
+    layoutTransform.translation.z = targetTransform.translation.z - (targetHeight != 0 ? clampValue(CAMERA_CLAMP_MIN, CAMERA_CLAMP_MAX, targetHeight * DEFAULT_LAYOUT_CAMERA_Z_OFFSET) : DEFAULT_LAYOUT_CAMERA_Z_OFFSET);
```
jennej: are you intentionally using height and z together here?  if so, can you add a comment about why this makes sense?

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -468,20 +470,27 @@ public void setSelectedExpression(Expression expression) {
     }
   }
 
+  private double clampValue(double min, double max, double val) {
+    return Math.min(max, Math.max(min, val));
```
jennej, line 486: It's a style choice, but I almost feel like you want a pictureClamp and an cameraClamp, which each have their own defaults inside, since using this function makes all your lines long and harder to read

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -468,20 +470,27 @@ public void setSelectedExpression(Expression expression) {
...
     double targetHeight = alignedBox.getHeight();
     double targetWidth = alignedBox.getWidth();
     double targetDepth = alignedBox.getDepth();
-    double defaultDistance = 1.0f;
     AffineMatrix4x4 targetTransform = targetImplementation.getAbsoluteTransformation();
     ClippedZPlane picturePlane = new ClippedZPlane();
 
     // Update layout camera marker
     AffineMatrix4x4 layoutTransform = AffineMatrix4x4.createIdentity();
+    double CAMERA_CLAMP_MAX = 100.0;
```
jennej: final :)

dabeshouse (commented):
I agree with Jen's suggestions and question.

More broadly these changes don't address the issue Melanie brought up around the refocus being too responsive, resetting any zoom a user might apply and shifting the view when it may not be desired. I will make that another issue and you can just regard this as a warning of work to come.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: yes, I've tried the depth value but the height value is more accurate in calculating the z-offset. I think that is because I'm offsetting the camera with the distance of a square.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: thx, define two new functions that use this clampValue function

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
weijchen: fix!

weijchen:
Fix Jen's suggestions. Will fix the issue Melanie brought up in another issue.

jennej (approved):
looks good

PR #287, opened 2022-11-21
Feature/sims icons

dabeshouse:


core/story-api/src/main/java/org/lgna/story/implementation/alice/AliceResourceUtilities.java
```diff
@@ -355,7 +354,7 @@ private static void findAndStoreResourceNames(ModelResource resource, String res
           visualName = enumToCamelCase(modelName.toString());
           textureName = arrayToEnum(splitName, i + 1, splitName.length);
           if (checkVisualAndTextureName(resource, visualName, textureName)) {
-            found = true;
+            Logger.warning("Initially failed to find resource names for '" + resource + "' and '" + resourceName + "' but did find for '" + modelName + "'");
             break;
           }
         }
```
jennej, line 360: this won't log at all if we loop through but don't find anything?

jennej (approved):
maybe add another warning, but otherwise seems good to me

core/story-api/src/main/java/org/lgna/story/implementation/alice/AliceResourceUtilities.java
dabeshouse: It still logs from `getModelNameFromClassAndResource`, called by `getModelNameFromClassAndResource`, this just reduces the redundant logging.

PR #288, opened 2022-11-23
T1017 Stretch images stored in textures before use to fix baby penguin

dabeshouse:
ImageUtilities.stretchToPowersOfTwo was created to handle user imported images. This applies it to model textures as well and fixes the baby penguin on Macs seeing this problem.


core/story-api/src/main/java/org/lgna/story/implementation/alice/AliceResourceUtilities.java
```diff
@@ -174,6 +177,16 @@ public static TexturedAppearance[] decodeTexture(URL url) {
     }
   }
 
+  // Stretch images to power of two to fix rendering on certain Mac graphics configurations.
+  // The problem was observed specifically with the Baby Penguin model.
+  private static void correctDimensions(TexturedAppearance ta) {
```
jennej, line 182: is it worth checking if the image is already a power of 2?

jennej (approved):

core/story-api/src/main/java/org/lgna/story/implementation/alice/AliceResourceUtilities.java
dabeshouse: This uses stretchToPowersOfTwo, which does that check first thing, so I did not repeat the test.

PR #294, opened 2022-12-24
some default values

jennej:
for TheAliceProject/Planning:540


core/story-api/src/main/java/org/lgna/story/IsVolumePreserved.java
```diff
@@ -64,6 +64,10 @@ private static boolean getValue(Object[] details, boolean defaultValue) {
     return defaultValue;
   }
 
+  public static Object getDefaultValue() {
```
dabeshouse, line 67: As a boolean, not an enum, this is not used. The menu items are added in BooleanFillerInner.
Consistent with the others, and hopeful about next steps, so I would say leave it.

Applies to others like the doubles in`TextScale`, `Duration`, and `AlongAxisOffset`. Let's be optimistic and bring them all in.

core/ide/src/main/java/org/alice/ide/croquet/models/cascade/StaticFieldAccessFillIn.java
```diff
@@ -61,26 +61,29 @@
...
+    return getInstance(value, false);
+  }
+  public static StaticFieldAccessFillIn getInstance(AbstractField value, boolean isDefault) {
     synchronized (map) {
       StaticFieldAccessFillIn rv = map.get(value);
-      if (rv != null) {
-        //pass
-      } else {
-        rv = new StaticFieldAccessFillIn(value);
+      if (rv == null) {
```
dabeshouse, line 73: Thank you. I can pass on // pass.

core/ide/src/main/java/org/alice/ide/croquet/models/cascade/ExpressionBlank.java
```diff
@@ -117,6 +117,7 @@ public <T> ExpressionBlank(Class<T> cls) {
   protected void updateChildren(List<CascadeBlankChild> children, BlankNode<Expression> blankNode) {
     IDE ide = IDE.getActiveInstance();
     if (ide != null) {
+      // this is less of an update and more of a 'create all the things from scratch'
```
dabeshouse, line 120: Sad but true. Worse still it seems to be triggered by side effect when checking `RtItem.isLast()`.

core/ide/src/main/java/org/alice/ide/croquet/models/ast/keyed/KeyedBlank.java
```diff
@@ -106,10 +108,8 @@ protected void updateChildren(List<CascadeBlankChild> children, BlankNode<JavaKe
               break;
             }
           }
-          if (isAlreadyFilledIn) {
-            //pass
-          } else {
-            children.add(JavaKeyedArgumentFillIn.getInstance(JavaMethod.getInstance(mthd)));
+          if (!isAlreadyFilledIn) {
```
dabeshouse, line 111: Clearer logic. Use of existing keyMethod variable. All good things.

core/ide/src/main/java/org/alice/ide/croquet/models/ast/keyed/KeyedBlank.java
```diff
@@ -92,6 +89,11 @@ private KeyedBlank(ArgumentListProperty<JavaKeyedArgument> argumentListProperty)
 
   @Override
   protected void updateChildren(List<CascadeBlankChild> children, BlankNode<JavaKeyedArgument> blankNode) {
+    // For clever reasons, this code has no idea what is in the scene, or what any of the methods can do.
+    // For this method (e.g. move or say), find the arguments it could have (e.g. duration, text, animation style)
+    // Once we know what the options are, we can to check if it already has a value (and thus is handled elsewhere)
```
dabeshouse: Extra "to" in "… we can to check…".

core/ide/src/main/java/org/alice/ide/cascade/fillerinners/ConstantsOwningFillerInner.java
```diff
@@ -83,12 +82,16 @@ private ConstantsOwningFillerInner(AbstractType<?, ?, ?> type) {
...
   @Override
   public void appendItems(List<CascadeBlankChild> items, ValueDetails<?> details, boolean isTop, Expression prevExpression) {
+    this.appendItems(items, details, isTop, prevExpression, null);
+  }
+
+  public void appendItems(List<CascadeBlankChild> items, ValueDetails<?> details, boolean isTop, Expression prevExpression, Object defaultValue) {
     AbstractType<?, ?, ?> type = this.getType();
     for (AbstractField field : type.getDeclaredFields()) {
       if (field.isPublicAccess() && field.isStatic() && field.isFinal()) {
-        //todo: should this be identical? to?
```
dabeshouse, line 89: Removing a ten year old todo that is of unclear intent. Thank you.

core/croquet/src/main/java/org/lgna/croquet/views/DropDown.java
```diff
@@ -179,11 +185,14 @@ public void paint(Graphics g) {
       if (isActive || DropDown.this.isInactiveFeedbackDesired()) {
         if (isActive) {
           if (buttonModel.isPressed()) {
+            // active
```
dabeshouse, line 188: Comments are a good addition. This logic could be simpler than how it is written.

Instead of nested ifs
Pressed -> gray127
Hover or inactive desired -> gray220
else background

Not asking you to change it, but I spent the thought to mentally untangle it, so I am jotting it down.

core/croquet/src/main/java/org/lgna/croquet/CascadeBlank.java
```diff
@@ -84,16 +85,10 @@ public final CascadeBlankChild[] getFilteredChildren(BlankNode<B> blankNode) {
...
-    for (int i = 0; i < N; i++) {
-      int index = N - i - 1;
-      if (isEmptySeparator(children.get(index))) {
-        children.remove(index);
-      } else {
-        break;
-      }
+    //remove separators at the end. there shouldn't be more than one, since we just removed all the duplicates.
+    final int last = children.size() - 1;
+    if (isEmptySeparator(children.get(last))) {
```
dabeshouse: Much simpler, but it breaks when the children are empty.
To test it add a field to the scene and try to set the initializer before setting the type. You will see the red queen.

core/ide/src/main/java/org/alice/ide/cascade/ExpressionCascadeManager.java
```diff
@@ -480,8 +470,19 @@ public void appendItems(List<CascadeBlankChild> items, BlankNode<Expression> bla
         }
       }
       if ((enumType != null) && this.areEnumConstantsDesired(enumType)) {
+        // default value- not all types have this method but we try. This approach only works for enums, since at this
+        // point in the code we only know what the desired type of the selection has to be, not what parameter we're
+        // attempting to fill.
+        Object defaultValue = null;
+        AbstractMethod mthd = type.getDeclaredMethod("getDefaultValue");
```
dabeshouse, line 477: Instead of this kind of reflection that only handles enums, it might be possible to change `ParamaterBlank` to not call `super.updateChildren()` and instead call a more specialized method on `ExpressionCascadeManager` like `appendItems` but with default in mind, or even in a parameter.
Let's merge this mostly as is, but maybe such an approach can get the next round of default values to show up?

core/ide/src/main/java/org/alice/ide/cascade/ExpressionCascadeManager.java
```diff
@@ -480,8 +470,19 @@ public void appendItems(List<CascadeBlankChild> items, BlankNode<Expression> bla
...
+        if (mthd != null && mthd.isStatic()) {
+          Object[] args = {};
+          VirtualMachine vm = StorytellingSceneEditor.getInstance().getVirtualMachine();
+          defaultValue = vm.ENTRY_POINT_invoke(null, mthd, args);
```
dabeshouse: The arg, `args`, is not needed and could be dropped.

core/ide/src/main/java/org/alice/ide/cascade/fillerinners/ConstantsOwningFillerInner.java
```diff
@@ -83,12 +82,16 @@ private ConstantsOwningFillerInner(AbstractType<?, ?, ?> type) {
...
+        boolean isDefault = defaultValue != null && field.getName().equals(defaultValue.toString());
```
dabeshouse, line 92: Could be computed inside the next `if` only when needed.
Then the two conditionals could collapse.

dabeshouse, line 92: Could be computed inside the next `if` only when needed.
Then the two conditionals could collapse.

dabeshouse (changes requested):
A fix for CascadeBlank is needed before merging. The rest is mostly comments.

core/ide/src/main/java/org/alice/ide/croquet/models/ast/keyed/KeyedBlank.java
jennej: ah, good catch. i re-wrote this one a few too many times

core/story-api/src/main/java/org/lgna/story/IsVolumePreserved.java
jennej: yeah, my optimism is showing for sure :)

core/croquet/src/main/java/org/lgna/croquet/CascadeBlank.java
jennej: oof, thanks.  i think i used all the braincells untangling the previous logic and didn't recover when writing mine.  I'll add a check

core/ide/src/main/java/org/alice/ide/cascade/fillerinners/ConstantsOwningFillerInner.java
jennej: ah, yes. i also like that better

dabeshouse (approved):

PR #297, opened 2023-01-11
Feature/migration fix

dabeshouse:


core/ast/src/main/java/org/lgna/project/Version.java
```diff
@@ -46,6 +46,7 @@
  * @author Dennis Cosgrove
  */
 public final class Version implements Comparable<Version> {
+  public static final Version VERSION_3_7 = new Version("3.7.0.0");
```
jennej, line 50: add a comment about why we need this variable so nobody has to hunt it down when they trip over this later?

jennej (approved):

core/ast/src/main/java/org/lgna/project/Version.java
dabeshouse: That is a much better idea than my instinct to start adding a version history here.

PR #302, opened 2023-01-24
Feature/sims update

dabeshouse:
Addresses internal issue 216 "Sims models do not reconnect joints correctly when changing resources"


jennej (approved):
thanks for the comments!

PR #303, opened 2023-01-24
Clear adapter element cache when program is stopped.

dabeshouse:
While debugging #302 I found the adapter factory has a cache that is not ever being cleared, and seems to not have much reused across runs, so this is a small change to clear it. More could be done


jennej (approved):
lgtm

PR #306, opened 2023-02-02
Feature/fov export

dabeshouse:
This is corrects the export of Camera fov and clipping plane export. It is in the IDE. It is in the tweedle library. It is in the player.
This is mainly to correct export. It should have been merged a few months back when the other changes went in

### DIFF
--- a/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/SymmetricPerspectiveCamera.java
+++ b/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/SymmetricPerspectiveCamera.java
@@ -55,22 +55,22 @@ import edu.cmu.cs.dennisc.math.property.AngleProperty;
 public class SymmetricPerspectiveCamera extends AbstractPerspectiveCamera {
   public final AngleProperty verticalViewingAngle = new AngleProperty(this, new AngleInRadians(0.5), true);
   public final AngleProperty horizontalViewingAngle = new AngleProperty(this, new AngleInRadians(Double.NaN), true);
-  private AngleInRadians effectiveHorizontalViewingAngle;
-  private AngleInRadians effectiveVerticalViewingAngle;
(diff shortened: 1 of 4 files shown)


core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/SymmetricPerspectiveCamera.java
```diff
@@ -54,4 +55,22 @@
 public class SymmetricPerspectiveCamera extends AbstractPerspectiveCamera {
   public final AngleProperty verticalViewingAngle = new AngleProperty(this, new AngleInRadians(0.5), true);
   public final AngleProperty horizontalViewingAngle = new AngleProperty(this, new AngleInRadians(Double.NaN), true);
+  private AngleInRadians effectiveHorizontalViewingAngle;
```
jennej: this feels nitpicky, but I feel like the getters and setters should use the same type?

jennej: and then i looked into the difference between AngleInRadians and Angle and i think i sprained something.

jennej (approved):

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/SymmetricPerspectiveCamera.java
dabeshouse: Not so picky. I have a mix of revolutions and radians along with wrapped and unwrapped values that are not always clear from the method or variable name.
I will see if I can make this a little clearer.

dabeshouse: I made it only a little clearer.

PR #308, opened 2023-03-03
Roll back overly tight, and wrong,  generic use for nebulous models

dabeshouse:
The generics broke export of sims models that were not persons.


jennej (approved):
👍

PR #313, opened 2023-04-11
Feature/exec java

dabeshouse:
Fix launching with `mvn exec:java -Dalice-ide`
mvn 3.3.1 adds use of project specific jvm.config files, so we will use it
The `-Dorg.alice.ide.rootDirectory` remains in the pom file where the relative location is correct. I found no permutation that worked in the top level jvm.config files.


jennej (approved):
lgtm

PR #320, opened 2023-05-25
Feature/cleanup

dabeshouse:
Some commits from the big VR branch that stand on their own. Together.


core/story-api/src/main/java/org/lgna/story/implementation/PerspectiveCameraMarkerImp.java
```diff
@@ -95,59 +92,63 @@ protected Color4f getDefaultMarkerColor() {
...
+    initialize();
+    sgVisuals = new Visual[] {createLensVisual(),
+        createCylinderVisual("Camera Cylinder 1", RADIUS),
+        createCylinderVisual("Camera Cylinder 2", RADIUS * 3),
+        createBoxVisual()};
 
-    this.sgAppearance = new SimpleAppearance();
-    this.sgAppearances = new SimpleAppearance[] {sgAppearance};
+    addLaserSightLine();
+    updateViewGeometry();
```
jennej, line 100: assuming you meant to drop setViewingAngle here?

jennej:
thank you for your service, removing all those unused empty classes.

jennej (approved):

core/story-api/src/main/java/org/lgna/story/implementation/PerspectiveCameraMarkerImp.java
dabeshouse: Yes. Those values are now set in initialize() and don't need to be set again.

dabeshouse: Hmmm… now it is never used. Guess I could delete it.

dabeshouse: Deleted it and a few related lines.

PR #321, opened 2023-05-31
Camera zoom

jennej:
I also would like to rename 

CameraMarkerTracker::markerToUpdate -> trackedMarker
CameraMarkerTracker::activeMarker -> selectedMarker

but held off to maybe do this after we resolve?


dabeshouse (changes requested):
A few things to consider.
- The zoom when focused may be larger than allowed. This means if the user tries the camera zoom arrows it can snap down to much closer. We could bump that max up, or change the logic to not snap back.
- Right click as selection is not quite there on the Mac and its weird mices. Using ctrl-click does select the thing, but the menu that pops up still applies to the last selected item. Using the two finger tap on a trackpad brings up the menu for the last selected but does not select the new item at all.
Otherwise looks good.
If you want to make the rename changes you suggest, they sound good to me, either now or later.

dportnoy15 (approved):

dabeshouse (approved):

PR #326, opened 2023-06-06
Feature/vr user models

dabeshouse:


core/story-api/src/main/java/org/lgna/story/implementation/StoryApiDirectoryUtilities.java
```diff
@@ -131,30 +132,24 @@ private static void askUserForModelGallery() {
...
+  }
+
+  private static File getDirectory(String name) {
     try {
       File installDirectory = getInstallDirectory();
       if (installDirectory != null) {
-        File starterProjectsDirectory = new File(installDirectory, STARTER_PROJECTS_NAME);
-        if (starterProjectsDirectory.isDirectory()) {
-          return starterProjectsDirectory;
+        File soundGalleryDirectory = new File(installDirectory, name);
```
jennej: soundGalleryDirectory looks like it snuck through here, swap this to a more generic version.

this is much cleaner now!

jennej (approved):
just the one rename and this is good

core/story-api/src/main/java/org/lgna/story/implementation/StoryApiDirectoryUtilities.java
dabeshouse: Oops. I will change that.

PR #339, opened 2023-07-21
Feature/center on camera

jennej:


core/ide/src/main/java/org/alice/stageide/sceneeditor/viewmanager/CameraMarkerTracker.java
```diff
@@ -180,7 +180,7 @@ private double clampCameraValue(double val) {
   private double clampPictureValue(double val) {
     // we intentionally allow a slightly larger max here, and the manipulator deals with it later.
     final double PICTURE_CLAMP_MAX = 100;
-    final double PICTURE_CLAMP_MIN = OrthographicCameraDragZoomManipulator.MIN_ZOOM;
```
dabeshouse, line 183: This changes leaves an unused import that triggers CheckStyle.

dabeshouse (changes requested):
Looks good, and it works, but is waiting on an import removal and another branch to be merged in.

PR #342, opened 2023-07-21
marker movement fix

dportnoy15:
Fix the issues where moving an object marker to an object or moving a camera marker to the main camera leaves a ghost


dabeshouse (commented):
After tracing where these two were common I have put up an alternative branch that notifies listeners when animations complete. Take a look and see what you think and we can pick one.

dabeshouse:
Alternative approach used in another branch

PR #345, opened 2023-07-26
Feature/camera drag

dabeshouse:


dportnoy15 (approved):

core/story-api/src/main/java/org/lgna/story/implementation/PerspectiveCameraMarkerImp.java
```diff
@@ -304,6 +304,10 @@ public Visual[] getSgVisuals() {
     return isVrActive ? sgVrVisuals : sgCameraVisuals;
   }
```
jennej, line 306: i have no idea what putInstance does.  might warrant a comment?

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
```diff
@@ -595,8 +596,9 @@ private void handleManipulatorSelection(SelectionEvent e) {
         } else {
           this.setSelectedField(field.getDeclaringType(), field);
         }
```
jennej, line 598: this function is starting to get scary, but it is also creeping into drag adaptor territory, so i think i'm backing away quietly to not wake the beast.

jennej (approved):

core/story-api/src/main/java/org/lgna/story/implementation/PerspectiveCameraMarkerImp.java
dabeshouse: It is a scary subsystem on every Element that I want to remove, but today it is useful.
A comment has been added to the code too.

core/ide/src/main/java/org/alice/stageide/sceneeditor/StorytellingSceneEditor.java
dabeshouse: So many dispatcher methods like this.
More "Uh oh" than "OO", amirite?

I join you in slowly stepping away.

PR #350, opened 2023-07-28
Feature/main camera visuals

dabeshouse:
Extends the handle_updates branch
Puts visuals directly on the main camera so they stay in sync instead of using a marker.


core/ide/src/main/java/org/alice/stageide/sceneeditor/interact/GlobalDragAdapter.java
```diff
@@ -684,13 +681,8 @@ public void undoRedoEndManipulation(AbstractManipulator manipulator, AffineMatri
       } else {
         UserField manipulatedField = sceneEditor.getFieldForInstanceInJavaVM(aliceThing);
         Group group = manipulatedField == null ? Application.DOCUMENT_UI_GROUP : Application.PROJECT_GROUP;
-        if (aliceThing instanceof SVRUser || aliceThing instanceof SCamera) {
-          undoOperation = new PredeterminedSetStartCameraTransformationActionOperation(group, false, this.getAnimator(), manipulatedField, originalTransformation, newTransformation, sceneEditor, manipulator.getUndoRedoDescription());
```
jennej, line 688: is PredeterminedSetStartCameraTransformationActionOperation still needed anywhere? can we jettison it for weight?

jennej (approved):

core/ide/src/main/java/org/alice/stageide/sceneeditor/interact/GlobalDragAdapter.java
dabeshouse: It was deleted in the same commit, but not much weight at 26 lines

PR #371, opened 2024-03-05
Extend Open for VR

dabeshouse:
Adds check to the file tab. Also adds thumbnail for current selection on that tab.


dabeshouse (commented):
Self reflection is good. Self code review, less so. But I am still merging.

PR #375, opened 2024-03-25
Code style- a passing attempt to make the code a bit easier to read

jennej:
I've attempted to ride the line between bananas and actually making a dent in the problem...


dabeshouse (approved):
A thousand times yes, or at least 235 times.

PR #382, opened 2024-04-02
Code style- QOL improvements

jennej:
please do sanity check my changes to hierarchy/component utilities, I went a little harder there than the rest.

### DIFF
--- a/core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
+++ b/core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
@@ -68,25 +68,6 @@ public class HierarchyUtilities {
     return true;
   }
 
-  private static <E extends AwtComponentView<?>> E getFirstToAccept(boolean isComponentACandidate, boolean isChildACandidate, boolean isGrandchildAndBeyondACandidate, AwtComponentView<?> component, Class<E> cls, Criterion<?>... criterions) {
-    assert component != null;
-    E rv = null;
-    if (isComponentACandidate && (cls == null || cls.isAssignableFrom(component.getClass())) && isAcceptedByAll(component, criterions)) {
-      rv = (E) component;
-    } else {
-      if (isChildACandidate && component instanceof AwtContainerView<?>) {
-        AwtContainerView<?> container = (AwtContainerView<?>) component;
-        for (AwtComponentView<?> componentI : container.getComponents()) {
-          rv = getFirstToAccept(isChildACandidate, isGrandchildAndBeyondACandidate, isGrandchildAndBeyondACandidate, componentI, cls, criterions);
-          if (rv != null) {
-            break;
-          }
-        }
-      }
-    }
-    return rv;
-  }
-
   private static <E extends AwtComponentView<?>> void updateAllToAccept(boolean isComponentACandidate, boolean isChildACandidate, boolean isGrandchildAndBeyondACandidate, java.util.List<E> list, AwtComponentView<?> component, Class<E> cls, Criterion<?>... criterions) {
     assert component != null;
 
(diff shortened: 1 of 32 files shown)


core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
```diff
@@ -57,42 +57,31 @@ private HierarchyUtilities() {
...
-        } else {
-          for (Criterion criterion : criterions) {
-            if (criterion.accept(component)) {
-              //pass
-            } else {
-              isAcceptedByAll = false;
-              break;
-            }
-          }
+  private static boolean checkAcceptance(AwtComponentView<?> component, Criterion<?>... criterions) {
```
dabeshouse: I have a preference for naming check functions like this with "is", perhaps "isAcceptedByAll".
Also, it is a carry over, but we could use the plural "criteria".

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
```diff
@@ -57,42 +57,31 @@ private HierarchyUtilities() {
...
+    boolean isAcceptedByAll = true;
```
dabeshouse: Could drop the variable and just return inside loop, or use streams to flatten further.

`return criteria == null || Arrays.stream(criteria).allMatch(criterion -> criterion.accept(component));`

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
dabeshouse, line 1: If you want to clean more here there are many unused methods and public methods with no external callers that could be made private.

dabeshouse (approved):
All good changes. Thank you.
Comments are optional further suggestions, or we can just merge and move on.

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
jennej: i'm more than happy to change to criteria, if i can do it in all 59 places criterions shows up...

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
jennej: I actually thought of changing that to a return right after I submitted.  I'll accept the fancy streams as well.

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
```diff
@@ -57,168 +57,41 @@ private HierarchyUtilities() {
...
-    HierarchyUtilities.updateAllToAccept(howMuch, list, component, cls, criterions);
+    HierarchyUtilities.updateAllToAccept(DEFAULT_HOW_MUCH.isComponentACandidate(), DEFAULT_HOW_MUCH.isChildACandidate(), DEFAULT_HOW_MUCH.isGrandchildAndBeyondACandidate(), list, component, cls, criteria);
     return list;
   }
 
-  public static <E extends AwtComponentView<?>> java.util.List<E> findAllMatches(AwtComponentView<?> component, Class<E> cls, Criterion<?>... criterions) {
-    return findAllMatches(component, DEFAULT_HOW_MUCH, cls, criterions);
-  }
-
   public static <E extends AwtComponentView<?>> java.util.List<E> findAllMatches(AwtComponentView<?> component, Class<E> cls) {
```
dabeshouse, line 94: This method can be deleted too and it will fall through to the method it calls, with the added benefit of always supplying an empty array of Criteria so the null checks are not needed.

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
```diff
@@ -57,168 +57,41 @@ private HierarchyUtilities() {
 
   public static final HowMuch DEFAULT_HOW_MUCH = HowMuch.COMPONENT_AND_DESCENDANTS;
```
dabeshouse, line 58: Not sure if you noticed, but this default translates to three true values and is never overridden.
Such flexibility we offer! - but do not use.

This is mostly true in ComponentUtilities as well, with a single exception that changes the first of the three values to true.

Apply this knowledge or not. Mostly I am just griping about it.

dabeshouse (approved):
A few ignorable comments from me.

core/croquet/src/main/java/org/lgna/croquet/views/HierarchyUtilities.java
jennej: i did look at that, and marvel at the options that we were offering but never using. it would definitely uncomplicate things further to remove all of it, i just wasn't sure if it could ever be useful?  because adding it back in would be  annoying.

PR #384, opened 2024-04-15
Feature/double click center

dabeshouse:
Listen for double click on tree in scene editor. When done in cameras other than main put that object in the center of the view for all non-main cameras.


dabeshouse:
Reviewed my own PR - trust me!
Merging in to get complete builds for testing.

core/ide/src/main/java/org/alice/stageide/sceneeditor/views/InstanceFactorySelectionPanel.java
```diff
@@ -310,15 +307,14 @@ protected void paintChildren(Graphics g) {
 
           Composite prevComposite = g2.getComposite();
```
jennej, line 309: there's no reason to create nextComposite here instead of on line 325 when it is set??

PR #385, opened 2024-04-22
Center for Main Camera

dabeshouse:


core/ide/src/main/java/org/alice/stageide/oneshot/OneShotUtilities.java
```diff
@@ -75,10 +76,10 @@ public static List<StandardMenuItemPrepModel> createMenuItemPrepModels(InstanceF
       ThisFieldAccessFactory thisFieldAccessFactory = (ThisFieldAccessFactory) instanceFactory;
       UserField field = thisFieldAccessFactory.getField();
       models.add(RenameFieldComposite.getInstance(field).getLaunchOperation().getMenuItemPrepModel());
+      AbstractType<?, ?, ?> fieldType = field.getValueType();
+      boolean isNotTargetingCamera = !fieldType.isAssignableTo(SCamera.class) && !fieldType.isAssignableTo(SVRUser.class);
```
jennej, line 80: isNotTargetingCamera can be final

core/ide/src/main/java/org/alice/stageide/oneshot/OneShotUtilities.java
```diff
@@ -87,12 +88,9 @@ public static List<StandardMenuItemPrepModel> createMenuItemPrepModels(InstanceF
       //      }
       models.add(RevertFieldOperation.getInstance(field).getMenuItemPrepModel());
 
-      if (!StorytellingSceneEditor.getInstance().isStartingCameraView()) {
-        if (!field.getValueType().isAssignableTo(SGround.class)) {
-          models.add(CenterCameraOnOperation.getInstance(field).getMenuItemPrepModel());
-        }
-      } else {
-        // TODO can we use the GetAGoodLookAt logic here to do something similar for the StartingCamera?
+      boolean cameraIsNotTargetingItself = !StorytellingSceneEditor.getInstance().isStartingCameraView() || isNotTargetingCamera;
```
jennej, line 91: final

jennej (approved):
the biggest thing I've noticed is that we're further splitting the actions in the scene editor from GetAGoodLookAtActionOperation, which might cause confusion/frustration with GetAGood... but that pretty easily creeps outside the scope of this review

core/ide/src/main/java/org/alice/stageide/oneshot/OneShotUtilities.java
dabeshouse: It could, but it would be one of very few (any?) final local variables in our code base, including 3 previous variables in this method that could be final.
Between the verbosity and the limited scope of a local variable, I am disinclined to use final for locals. I am going to mull further and let this sit for another day.

I think Java picked the wrong default. We should have to declare that a variable can be changed.

dabeshouse:
> the biggest thing I've noticed is that we're further splitting the actions in the scene editor from GetAGoodLookAtActionOperation, which might cause confusion/frustration with GetAGood... but that pretty easily creeps outside the scope of this review

You make a good point. This does leave us with two very similar methods with slightly different behavior, accessibility (double click, menu, code), and one of them varies results based on where the camera starts.

I am going to play a little more with both and think this over.

PR #387, opened 2024-04-26
Icon factory cleanups

jennej:
just me crusading against poor innocent files that haven't actually done anything wrong


core/ide/src/main/java/org/alice/stageide/gallerybrowser/shapes/BillboardDragModel.java
```diff
@@ -42,23 +42,22 @@
...
 import org.alice.stageide.modelresource.ResourceNode;
 import org.lgna.croquet.SingleSelectTreeState;
 import org.lgna.croquet.Triggerable;
-import org.lgna.croquet.icon.IconFactory;
-import org.lgna.story.EmployeesOnly;
 import org.lgna.story.SBillboard;
 
 import java.util.UUID;
 
+// generic drag models.
```
dabeshouse: Was including this comment an oversight?

core/ide/src/main/java/org/alice/stageide/icons/ShapeIconFactory.java
```diff
@@ -42,28 +42,38 @@
...
-  }
 
-  public static JointIconFactory getInstance() {
-    return SingletonHolder.instance;
-  }
+public class ShapeIconFactory extends ResolutionIndependentIconFactory {
+  Constructor<? extends ShapeIcon> iconConstructor;
 
-  private JointIconFactory() {
+  public ShapeIconFactory(Class<? extends ShapeIcon> cls)  {
```
dabeshouse, line 58: I was starting to suggest replacing the Class with a direct reference to the new method, but once I started down that path I saw how much more of the icon factory hierarchy can be collapsed or eliminated.
I think I will offer it as possible next step PR.

To be overly precise, the type would become `Function<Dimension, ShapeIcon>`, called with things like `DiscIcon::new` and invoked with `iconConstructor.apply(size)`.

core/ide/src/main/java/org/alice/stageide/icons/ShapeIconFactory.java
```diff
@@ -42,28 +42,38 @@
...
-public class JointIconFactory extends ResolutionIndependantIconFactory {
-  private static class SingletonHolder {
-    private static JointIconFactory instance = new JointIconFactory();
-  }
 
-  public static JointIconFactory getInstance() {
-    return SingletonHolder.instance;
-  }
+public class ShapeIconFactory extends ResolutionIndependentIconFactory {
+  Constructor<? extends ShapeIcon> iconConstructor;
```
dabeshouse, line 56: Could be final ;)

dabeshouse (approved):

core/ide/src/main/java/org/alice/stageide/gallerybrowser/shapes/BillboardDragModel.java
jennej: oops, yeah 😅

core/ide/src/main/java/org/alice/stageide/icons/ShapeIconFactory.java
jennej: I support all ideas to dismantle the icon factory industrial complex!

PR #388, opened 2024-04-29
Feature/person init

dabeshouse:


core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/IngredientsComposite.java
```diff
@@ -809,6 +777,38 @@ public PersonResource createResourceFromStates() {
...
+    Outfit fullbody = getFullBodyOutfitState().getValue();
+    if ((!topsAndBottomsAvailable || lastActiveOutfitTab != topAndBottomTab) && fullbody != null) {
+      return fullbody;
+    }
+
+    final TopPiece top = getTopPieceState().getValue();
+    final BottomPiece bottom = getBottomPieceState().getValue();
+    return switch (lifeStage) {
+      // TODDLER does not have tops and bottom options so this should not be reached. Here to cover all cases.
+      case TODDLER -> fullbody;
```
jennej, line 792: can fullbody be null here? do we care for this case that should not be reached?

jennej (approved):

PR #393, opened 2024-05-03
Use the same Icon for the camera

jennej:
we use this icon for the camera when we're seeing it as an object in the scene, not to be confused with the icon we use for the camera when we know it is the camera

resolves #770


dabeshouse (approved):
The new camera icon is 128x128.
This matches three of the four existing icons in the 256x256 folder, so this looks good to me.

PR #398, opened 2024-06-14
Feature/scene letterboxing

jennej:
this is just the code cleanup preamble, but it was getting big.  The actual letterboxing is coming soon


dabeshouse (approved):
So much "code" goes poof.

PR #405, opened 2024-07-25
Feature/update jogl

dabeshouse:
No longer need our branch build. Back to public builds.
Dropping 32 bit builds.


jennej (approved):
we're in The Future

PR #406, opened 2024-07-28
Cleaning up the transformation code

jennej:
_New and _AffectReturnValuePassedIn have been almost entirely eradicated from the codebase

I would like to go a bit further with the PicturePlaneUtils functions, but this was getting massive, and it seems like some of it might happen naturally if I dive deeper into the pose code, so I left things as they were.


core/scenegraph/src/main/java/edu/cmu/cs/dennisc/render/PicturePlaneUtils.java
```diff
@@ -157,185 +155,38 @@ private static Vector4 transformFromCameraToViewport_AffectReturnValuePassedIn(V
...
+    final Rectangle actualViewport = renderTarget.getActualViewportAsAwtRectangle(sgCamera);
+    transformFromCameraToViewport_AffectReturnValuePassedIn(rv, renderTarget, sgCamera, actualViewport);
     return rv;
   }
 
-  public static Vector4 transformFromViewportToCamera(Vector4 rv, Vector4 xyzw, RenderTarget renderTarget, AbstractCamera sgCamera) {
-    rv.set(xyzw);
+  public static Vector4 transformFromViewportToCamera(Vector4 xyzw, RenderTarget renderTarget, AbstractCamera sgCamera) {
+    Vector4 rv = new Vector4(xyzw);
     return transformFromViewportToCamera_AffectReturnValuePassedIn(rv, renderTarget, sgCamera);
```
dabeshouse: This, and others here, still have the odd suffixes. They are all private, so I can see where you drew the line.
But this one, and some of the others, have only one caller and could be inlined and removed.
That can be for a next round of fixes.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/AbstractCamera.java
```diff
@@ -43,41 +43,13 @@
...
 import edu.cmu.cs.dennisc.property.InstanceProperty;
-import edu.cmu.cs.dennisc.render.RenderTarget;
-
-import java.awt.Point;
 
 /**
  * @author Dennis Cosgrove
  */
 public abstract class AbstractCamera extends Leaf {
-  public Point transformToAWT(Point rv, Vector4 xyzw, RenderTarget renderTarget) {
```
dabeshouse, line 58: Having these on camera (or perhaps RenderTarget) makes some sense.
Having the logic bounce up through Component and out to static helper methods, less so, so I can see why you took them out.
If you are looking for a non-static home for some of the functions, these might return - with implementations not passthroughs.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/util/TransformationUtilities.java
dabeshouse, line 1: The remaining `public static` methods are called from Component and perhaps should be methods there.
This might happen in this PR, another PR, or not at all.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/render/PicturePlaneUtils.java
```diff
@@ -55,22 +53,22 @@
...
  * @author Dennis Cosgrove
  */
 public class PicturePlaneUtils {
-  private static Vector4 transformFromAWTToViewport(Vector4 rv, Point p, double z, Rectangle actualViewport) {
-    rv.x = p.x;
-    rv.y = actualViewport.height - p.y;
-    rv.z = z;
-    rv.w = 1.0;
-    return rv;
+  private static Vector4 transformFromAWTToViewport(Point p, double z, Rectangle actualViewport) {
```
dabeshouse: This has no callers. Do we need it?

dabeshouse (approved):
There are more changes that could be made.
They could be put in this PR, but don't need to be.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/render/PicturePlaneUtils.java
jennej: Yeah, this is the stuff I'm hoping I can sort out/possibly eliminate with the pose code work. I have them gone in a stash.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/util/TransformationUtilities.java
jennej: oh, i was in too deep to see this, thank you.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Component.java
```diff
@@ -238,37 +237,107 @@ protected void fireHierarchyChanged() {
...
-    return TransformationUtilities.transformTo(xyz, this, to);
-  }
 
   public Vector3 transformFrom(Vector3 xyz, Component from) {
     return from.transformTo(xyz, this);
   }
 
+  private boolean isAbsolute() {
+    //todo: call isSceneOf()?
+    if ((this instanceof Scene) || ((ReferenceFrame) this == AsSeenBy.SCENE)) {
```
dabeshouse: Now that this is on Component the ReferenceFrame is more obviously not the case, so it simplifies to `instanceOf`, or just false with an overriding true in Scene.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Component.java
```diff
@@ -258,114 +234,93 @@ protected void fireHierarchyChanged() {
...
-
-  public Vector3 transformTo_New(Vector3 xyz, Component to) {
-    return TransformationUtilities.transformTo_New(xyz, this, to);
-  }
-
-  public Vector3 transformFrom_New(Vector3 xyz, Component from) {
-    return from.transformTo_New(xyz, this);
+  // Vector4
+  public Vector4 transformToAbsolute(Vector4 xyz) {
+    Vector4 rv = new Vector4(xyz);
```
dabeshouse, line 239: Could be inside conditional. No need to make a new one if it is the same.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Component.java
```diff
@@ -258,114 +234,93 @@ protected void fireHierarchyChanged() {
...
+      AffineMatrix4x4 m = getAbsoluteTransformation();
+      m.transform(rv);
+    }
+    return rv;
   }
 
-  public Vector4 transformTo_AffectReturnValuePassedIn(Vector4 xyzw, Component to) {
-    return TransformationUtilities.transformTo_AffectReturnValuePassedIn(xyzw, this, to);
+  public Vector4 transformFromAbsolute(Vector4 xyz) {
+    Vector4 rv = new Vector4(xyz);
```
dabeshouse, line 248: As above, this could be inside the conditional. And a few more below…

dabeshouse (approved):
lgtm
If you are looking for more, I pointed out minor Component tweaks, or they can be in the next round, or not.

PR #409, opened 2024-08-04
Fix bug #756 and improve function sorting a bit

dportnoy15:
Avoid hiding property accessors under some groupings, sort functions by their display name instead of their Java name, and sort functions alphabetically under each return type

### DIFF
--- a/core/ide/src/main/java/org/alice/ide/member/FunctionTabComposite.java
+++ b/core/ide/src/main/java/org/alice/ide/member/FunctionTabComposite.java
@@ -44,11 +44,13 @@ package org.alice.ide.member;
 
 import edu.cmu.cs.dennisc.java.util.Lists;
 import edu.cmu.cs.dennisc.java.util.Maps;
+
 import org.alice.ide.IDE;
 import org.alice.ide.croquet.codecs.StringCodec;
 import org.alice.ide.croquet.models.ui.preferences.IsEmphasizingClassesState;
 import org.alice.ide.instancefactory.InstanceFactory;
 import org.alice.ide.member.views.FunctionTabView;
+
 import org.lgna.croquet.ImmutableDataSingleSelectListState;
 import org.lgna.project.ast.AbstractMethod;
 import org.lgna.project.ast.AbstractType;
@@ -103,11 +105,25 @@ public final class FunctionTabComposite extends MemberTabComposite<FunctionTabVi
     if (instanceFactory != null) {
       AbstractType<?, ?, ?> type = instanceFactory.getValueType();
       while (type != null) {
-        for (AbstractMethod method : type.getDeclaredMethods()) {
-          AbstractType<?, ?, ?> returnType = method.getReturnType();
-          if (returnType != JavaType.VOID_TYPE && isInclusionDesired(method)) {
-            List<AbstractMethod> methods = funcByType.computeIfAbsent(returnType, k -> Lists.newLinkedList());
-            methods.add(method);
+        if (type instanceof NamedUserType) {
+          NamedUserType namedUserType = (NamedUserType) type;
+
+          UserMethodsSubComposite userMethodsSubComposite = this.getUserMethodsSubComposite(namedUserType);
+
+          for (AbstractMethod method : userMethodsSubComposite.getMethods()) {
+            AbstractType<?, ?, ?> returnType = method.getReturnType();
+            if (returnType != JavaType.VOID_TYPE && isInclusionDesired(method)) {
+              List<AbstractMethod> methods = funcByType.computeIfAbsent(returnType, k -> Lists.newLinkedList());
+              methods.add(method);
+            }
+          }
+        } else if (type instanceof JavaType) {
+          for (AbstractMethod method : type.getDeclaredMethods()) {
+            AbstractType<?, ?, ?> returnType = method.getReturnType();
+            if (returnType != JavaType.VOID_TYPE && isInclusionDesired(method)) {
+              List<AbstractMethod> methods = funcByType.computeIfAbsent(returnType, k -> Lists.newLinkedList());
+              methods.add(method);
+            }
           }
         }
         type = type.isFollowToSuperClassDesired() ? type.getSuperType() : null;
@@ -115,11 +131,19 @@ public final class FunctionTabComposite extends MemberTabComposite<FunctionTabVi
     }
 
     List<AbstractType<?, ?, ?>> types = Lists.newArrayList(funcByType.keySet());
+    List<MethodsSubComposite> rv = Lists.newArrayList();
+    UnclaimedMethodsComposite methodsComposite = this.getUnclaimedMethodsComposite();
+
     types.sort(IDE.getActiveInstance().getApiConfigurationManager().getTypeComparator());
-    List<MethodsSubComposite> rv = Lists.newArrayListWithInitialCapacity(types.size());
+
     for (AbstractType<?, ?, ?> type : types) {
       FunctionsOfReturnTypeSubComposite subComposite = FunctionsOfReturnTypeSubComposite.getInstance(type);
-      subComposite.setMethods(funcByType.get(type));
+
+      // sort methods under each return type
+      List<AbstractMethod> typeMethods = funcByType.get(type);
+      methodsComposite.sortAndSetMethods(typeMethods);
+
+      subComposite.setMethods(typeMethods);
       rv.add(subComposite);
     }
     return rv;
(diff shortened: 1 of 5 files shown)


dportnoy15:
I also looked a lot into how sorting is done for each of the different groupings, and would like to discuss it to see whether any of it is worth changing

core/ide/src/main/java/org/alice/ide/member/MemberTabComposite.java
```diff
@@ -197,20 +199,34 @@ private List<MethodsSubComposite> getSubCompositesAlphabetically() {
...
-        for (AbstractMethod method : type.getDeclaredMethods()) {
-          if (this.isAcceptable(method)) {
-            if (isInclusionDesired(method)) {
-              methods.add(method);
+        if (type instanceof NamedUserType) {
+          NamedUserType namedUserType = (NamedUserType) type;
+
+          UserMethodsSubComposite userMethodsSubComposite = this.getUserMethodsSubComposite(namedUserType);
+          for (AbstractMethod method : userMethodsSubComposite.getMethods()) {
+            if (this.isAcceptable(method)) {
```
jennej: this code is similarly repeaty, and also these two checks can be combined with &&

core/ide/src/main/java/org/alice/ide/member/FunctionTabComposite.java
```diff
@@ -103,23 +105,45 @@ private List<MethodsSubComposite> getByReturnTypeSubComposites() {
...
-        for (AbstractMethod method : type.getDeclaredMethods()) {
-          AbstractType<?, ?, ?> returnType = method.getReturnType();
-          if (returnType != JavaType.VOID_TYPE && isInclusionDesired(method)) {
-            List<AbstractMethod> methods = funcByType.computeIfAbsent(returnType, k -> Lists.newLinkedList());
-            methods.add(method);
+        if (type instanceof NamedUserType) {
+          NamedUserType namedUserType = (NamedUserType) type;
+
+          UserMethodsSubComposite userMethodsSubComposite = this.getUserMethodsSubComposite(namedUserType);
+
```
jennej: this is long and repeat-y  maybe move lines 113-119/121-127 into a separate function?

jennej (changes requested):

core/ide/src/main/java/org/alice/ide/member/MemberTabComposite.java
dabeshouse: Every use of `isAcceptable` is followed by `isInclusionDesired`. (But not the other way)
Not sure if having the former call the latter simplifies anything, since it would require adding `isAcceptable` or calls to super to two methods.

core/ide/src/main/java/org/alice/ide/member/FunctionTabComposite.java
dabeshouse: Repeaty, yes.
It bothers me that you needed to get the methods from an AbstractType in a different way for JavaType and NamedUserType, but I can see they do produce different results.

I would go as far as pushing 107-130 onto AbstractType and its subclasses. Have the type walk its own hierarchy to return the list of methods, then consider that list here, and above in Members.

dabeshouse:
> I also looked a lot into how sorting is done for each of the different groupings, and would like to discuss it to see whether any of it is worth changing

Having arbitrary ordering within groups is a little unpleasant. They could be secondarily alpha ordered.

Joints ordered by hierarchy would be nice to have, but that may be too much of a pain.

core/ide/src/main/java/org/alice/ide/member/FunctionTabComposite.java
dportnoy15: I'd originally taken that from getSubCompositesByCategory and put it into the other two getGetSubcomposites functions without changing it too much. I was considering splitting it off somehow to reduce repetition as well. Will see about moving it into AbstractType and subclasses.

dportnoy15:
Here's what I found about the sorting order for functions (specifically when selecting the Scene, i.e. "this"):

Sort Alphabetically:
 - obviously, sorted alphabetically, although I changed it to use display name instead of java name for the sort

Sort Ny Return Type:
 - it currently sorts the return types themselves using a hard-coded order, with java types first, followed the Tweedle (?) types, such as SThing, etc. (see StoryApiConfigurationManager.TypeComparator)
   - This probably makes sense, but we could change it if we want, esp. since users who don't know Java might be confused about the sort
 - I made it sort alphabetically (using the display name) within return types

Sort By Category:
 - the sort order for categories is:
      1. the Scene's editable functions
          - methods are listed first, followed by field getters and setters, see UserMethodsSubComposite.getMethods()
          - getSubCompositesByCategory just returns the UserMethodsSubComposite instance, without calling getMethods() itself
          - We could call getMethods() directly and create a new SubComposite, if we want to change this sort order
      2. a hard-coded list of categories, whose order is defined by categoryFunctionSubComposites and     categoryOrAlphabeticalFunctionSubComposites in StoryApiConfigurationManager
          - The per-category order is defined (mostly) by a hard-coded list of function names in the class constructors for the classes associated with these categories, e.g. AtmosphereFunctionsComposite
          - Looks like Joint functions (the only composite in categoryOrAlphabeticalFunctionSubComposites) are alpha sorted
      3. "other" methods, which are alpha sorted by getName(). would be easy to also change to use display name

THere's also probably parallel logic for sorting procedures, but I didn't look at that

dabeshouse:
Thanks for the detailed descriptions.

I think aim for using display name alpha ordering within all groupings in both functions and procedures, and address the code review suggestions.

No need for other reorderings, either for pedagogical reasons, or what we might think is clarifying.

core/ide/src/main/java/org/alice/ide/member/MemberTabComposite.java
```diff
@@ -263,9 +261,31 @@ private List<MethodsSubComposite> getSubCompositesByCategory() {
...
   }
 
+  protected List<? extends AbstractMethod> getAcceptableMethodsForType(AbstractType<?, ?, ?> type) {
+    List<AbstractMethod> methods = new LinkedList<>();
+
+    if (type instanceof NamedUserType) {
+      NamedUserType namedUserType = (NamedUserType) type;
+
+      UserMethodsSubComposite userMethodsSubComposite = this.getUserMethodsSubComposite(namedUserType);
+      System.out.println("CHECK type: " + namedUserType.name + "  num methods: " + userMethodsSubComposite.getMethods().size());
```
dabeshouse: Was this left in from debugging?
It should probably be removed.

core/ide/src/main/java/org/alice/ide/member/MemberTabComposite.java
```diff
@@ -231,29 +236,22 @@ private List<MethodsSubComposite> getSubCompositesByCategory() {
...
         } else if (type instanceof JavaType) {
-          JavaType javaType = (JavaType) type;
-          for (JavaMethod javaMethod : javaType.getDeclaredMethods()) {
-            if (this.isAcceptable(javaMethod)) {
-              if (isInclusionDesired(javaMethod)) {
-                javaMethods.add(javaMethod);
-              }
-            }
-          }
+          javaMethods.addAll((List<JavaMethod>) getAcceptableMethodsForType(type));
```
dabeshouse, line 239: Why did `getAcceptableMethodsForType` only replace for `JavaType`, but not `NamedUserType` just above?

core/ide/src/main/java/org/alice/ide/member/UnclaimedMethodsComposite.java
```diff
@@ -51,7 +54,18 @@
  * @author Dennis Cosgrove
  */
 public abstract class UnclaimedMethodsComposite extends FilteredMethodsSubComposite {
-  private final Comparator<AbstractMethod> comparator = FilteredMethodsSubComposite::compareMethodNames;
+  private final Comparator<AbstractMethod> comparator = UnclaimedMethodsComposite::compareMethodNames;
+
+  protected static int compareMethodNames(AbstractMethod methodA, AbstractMethod methodB) {
+    if (methodA == null) {
```
dabeshouse: I mostly just want to share what I recently found, but these null checks can be dropped if you define the comparator above as `comparator = Comparator.nullsLast(UnclaimedMethodsComposite::compareMethodNames)`.
This is a simplification we could apply elsewhere.

dabeshouse (changes requested):
I am resisting comments that are more about older code and would just lead to scope creep, but I found a few small things for you to consider.

dabeshouse (approved):

PR #410, opened 2024-08-05
Checkstyle would like the class name to match the file name

jennej:


dabeshouse (approved):
Thank you.
This should have happened a while back.

PR #411, opened 2024-08-05
final tweak (for now) on the icon factories

jennej:
a continuation of https://github.com/TheAliceProject/alice3/pull/387, apparently i got 90% done with this and then it got stashed instead of finished


dabeshouse (approved):
Looks like good steps in the right direction.

PR #412, opened 2024-08-12
Feature/universal neb build

dabeshouse:


core-nonfree/story-api-nonfree/src/main/java/edu/cmu/cs/dennisc/nebulous/Manager.java
```diff
@@ -82,15 +82,21 @@ public class Manager {
...
       //throw new RuntimeException( lre );
     } catch (Throwable t) {
-      JOptionPane.showMessageDialog(null, "failed to initialize art assets");
-      t.printStackTrace();
+      // Only show the dialog if it was not shown in the last second.
+      if (System.currentTimeMillis() - lastNotifiedAboutAssets > 1000) {
+        JOptionPane.showMessageDialog(null, "failed to initialize art assets");
+        t.printStackTrace();
+      }
+      lastNotifiedAboutAssets = System.currentTimeMillis();
```
jennej: by updating the time outside this loop, this will only show the message when it has been more than a second since we last attempted to show the message.

it is entirely possible that this is the effect you want? these errors appear in batches, or you just want to see the message once and this achieves that.  It just isn't what your comment says, so one or the other should be changed..

core-nonfree/story-api-nonfree/src/main/java/edu/cmu/cs/dennisc/nebulous/Manager.java
```diff
@@ -82,15 +82,21 @@ public class Manager {
 
   public static native void setDebugDraw(boolean debugDraw);
 
+  private static long lastNotifiedAboutAssets;
```
jennej: should this be initialized?

core-nonfree/story-api-nonfree/src/main/java/edu/cmu/cs/dennisc/nebulous/Manager.java
dabeshouse: I am not sure what loop you mean.
I update the value with the most recent time the error has happened or when the user has dismissed the dialog notifying the error has happened.
The intent it to not see the dialog once, and not dozens of times, when initialization is failing. It was useful when testing builds and is a minor kindness for users if there is a system problem.

core-nonfree/story-api-nonfree/src/main/java/edu/cmu/cs/dennisc/nebulous/Manager.java
dabeshouse: The default value for a long is 0 so I did not.

core-nonfree/story-api-nonfree/src/main/java/edu/cmu/cs/dennisc/nebulous/Manager.java
jennej: i don't know why i said loop. i meant inside the conditional

dabeshouse: Updated the comment to clarify.

core-nonfree/story-api-nonfree/src/main/java/edu/cmu/cs/dennisc/nebulous/Manager.java
dabeshouse: But a quick review of Alice code shows initializing to 0 is more frequent, so I have joined in.

PR #414, opened 2024-08-29
Less employees only

dabeshouse:
Code cleanup. Almost all gone, but two more to sort out.


core/story-api/src/main/java/org/lgna/story/Color.java
```diff
@@ -76,22 +78,38 @@ public Color(@ValueTemplate(detailsEnumCls = PortionDetails.class) Number red, @
...
   }
 
+  public static Color fromProperty(Color4fProperty colorProp) {
+    return createInstance(colorProp.getValue());
+  }
+
+  public Color interpolateTo(Color target, double portion) {
+    return createInstance(Color4f.createInterpolation(toColor4f(), target.toColor4f(), (float) portion));
+  }
+
```
jennej, line 88: do we want to do any defensive copy-ing of things to prevent out internal bits from being exposed?

core/story-api/src/main/java/org/lgna/story/Key.java
```diff
@@ -252,12 +252,11 @@ private Key(int keyCode) {
     MapContainer.map.put(this.keyCode, this);
   }
```
jennej, line 254: this one is safe, because it is an int, right?

core/story-api/src/main/java/org/lgna/story/Position.java
```diff
@@ -50,28 +50,18 @@
 public final class Position {
   private final Point3 internal;
```
jennej, line 52: another place we could add a copy

core/story-api/src/main/java/org/lgna/story/Orientation.java
```diff
@@ -51,7 +51,7 @@
 public final class Orientation {
   private final OrthogonalMatrix3x3 internal;
```
jennej, line 54: another place we could add a copy

core-nonfree/story-api-nonfree/src/main/java/org/lgna/story/SRoom.java
```diff
@@ -71,10 +71,6 @@ protected Texture create() {
     private final String textureKey;
   }
```
jennej, line 73: now we will just have to wonder ;)

core/story-api/src/main/java/org/lgna/story/Orientation.java
```diff
@@ -63,7 +63,7 @@ public Orientation(Number x, Number y, Number z, Number w) {
     this(new OrthogonalMatrix3x3(new UnitQuaternion(x.doubleValue(), y.doubleValue(), z.doubleValue(), w.doubleValue())));
   }
```
jennej, line 67: instead of exposing this, what if we added createUnitQuaternion and createEulerAngles as functions on Orientation?

jennej (commented):
So many things are now in much better places!

core-nonfree/story-api-nonfree/src/main/java/org/lgna/story/SRoom.java
dabeshouse: Wonder if it is valid? Undoubtedly.

But of course it is the existence of the test that creates doubt.

If I tell you "It's raining outside.", you will likely grab an umbrella.
If I say "Undoubtedly, it is raining outside.", you are more likely to find a window and check.

core/story-api/src/main/java/org/lgna/story/Color.java
dabeshouse: A disturbing thought occurs to me. What if some of these arguments are supposed to reference the same object?
It may be that something will break if we make copies.

core/story-api/src/main/java/org/lgna/story/Orientation.java
dabeshouse: Expose?
This was unchanged and already public. I did make the constructor above (line 54) public, but that one just takes the internal matrix.
`UnitQuaternion` has a method `createOrthogonalMatrix3x3()` that could be used here instead of the wrapping constructor call, but it seems kind of pointless.

core/story-api/src/main/java/org/lgna/story/Color.java
jennej: Disturbing indeed, but possible

core/story-api/src/main/java/org/lgna/story/Orientation.java
jennej: github showed this comment in the most useless place. i'm referencing the change on line 66 where getInternal is made public

core/story-api/src/main/java/org/lgna/story/Key.java
dabeshouse: Yeah, if the int value changes there is something bigger to worry about.

core/story-api/src/main/java/org/lgna/story/Orientation.java
dabeshouse: Done. I am not crazy about the "create…" notation, but it is established.

core/story-api/src/main/java/org/lgna/story/Color.java
dabeshouse: The two below do concern mutable values, but Color4f does not. It has 4 final values for r, g, b, and alpha.

core/story-api/src/main/java/org/lgna/story/Position.java
dabeshouse: Added note in code.

core/story-api/src/main/java/org/lgna/story/Orientation.java
dabeshouse: Added note in code.

jennej (approved):

PR #418, opened 2024-09-19
Feature/javafx nbm

dabeshouse:


netbeans/src/main/java/org/alice/netbeans/project/ProjectCodeGenerator.java
```diff
@@ -131,7 +129,7 @@ public static Collection<FileObject> generateCode(File aliceProjectFile, File ja
       }
     }
 
-    if (progressHandle != null) {
+      if (progressHandle != null) {
```
dportnoy15: Probably don't want this extra whitespace

netbeans/src/main/java/org/alice/netbeans/project/ProjectCodeGenerator.java
```diff
@@ -157,7 +153,7 @@ public static Collection<FileObject> generateCode(File aliceProjectFile, File ja
         }
       }
 
-      TextFileUtilities.write(file, code);
+        TextFileUtilities.write(file, code);
```
dportnoy15: Extra whitespace again

netbeans/src/main/java/org/alice/netbeans/project/ProjectCodeGenerator.java
```diff
@@ -204,4 +204,33 @@ public void run() {
     }
     return filesToOpen;
   }
+
+  private static FileObject generateLauncher(File javaSrcDirectory) {
```
dportnoy15, line 208: I'm wondering if it makes sense to also pass the filename in as a parameter, although it's not something I feel strongly about.

dportnoy15 (changes requested):

netbeans/src/main/java/org/alice/netbeans/project/ProjectCodeGenerator.java
dabeshouse: Oops. Thank you.

netbeans/src/main/java/org/alice/netbeans/project/ProjectCodeGenerator.java
dabeshouse: Oops and thank you again.

netbeans/src/main/java/org/alice/netbeans/project/ProjectCodeGenerator.java
dabeshouse: Since the file name must match the file content I think it belongs inside the method as well.
I made it static and final to match.

dportnoy15 (approved):

jennej (approved):

PR #419, opened 2024-10-26
some render target/drag adapter cleanups

jennej:
started with me trying not to get too distracted by the weirdness in the IK drag adapter world, and then some other small changes


core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/GlrRenderTarget.java
```diff
@@ -367,7 +359,6 @@ public Point transformFromCameraToAWT(Vector4 xyzw, AbstractCamera sgCamera) {
   public  Point transformFromCameraToAWT(Point3 xyzw,  AbstractCamera sgCamera) {
     return transformFromCameraToAWT(new Vector4(xyzw.x, xyzw.y, xyzw.z, 1.0), sgCamera);
   }
-  private final GlrRenderFactory glrRenderer;
```
dabeshouse, line 370: I understand removing this, since most other uses are already references to the static singleton anyway, and that singleton is what this always is. But it still bugs me to add more references when I have a general goal to eliminate static singletons whenever possible.
Eh, at least it is now more uniform and not making an extra step through RenderUtils.

dabeshouse (approved):
Less code. Helpful deprecation comments. lgtm
(My one comment is just a comment)

PR #421, opened 2024-11-03
awt vs viewport cleanups

jennej:
This started with me trying to understand why the pose tool was the only thing using viewport coordinates, and then we spent some time yak grooming.


core/story-api/src/main/java/org/lgna/story/event/MouseClickEvent.java
```diff
@@ -55,20 +55,4 @@ public abstract class MouseClickEvent extends AbstractEvent {
   public MouseClickEvent(MouseClickEventImp e) {
     this.implementation = e;
   }
-
```
dabeshouse: These removals leave unused imports here and in the corresponding Imp class that trigger CheckStyle.
(I should figure out how to automate the fixing rather than the complaining. Maybe move this to a GitHub action?)

dabeshouse (changes requested):
lgtm, modulo StyleCheck on imports

core/story-api/src/main/java/org/lgna/story/event/MouseClickEvent.java
```diff
@@ -55,20 +55,4 @@ public abstract class MouseClickEvent extends AbstractEvent {
...
-  public Double getScreenDistanceFromLeft() {
```
dabeshouse: These should not be deleted. They are used by reflection.
Without them the event values are not available
image

jennej:
i removed the final commit, which was causing all the problems :)

dabeshouse (approved):

PR #423, opened 2024-11-27
feature/negative-torus-radii

dportnoy15:
Fixes issue #702: Trying to set the sizes of models or different shapes to be negative will instead shrink them down to just above 0 size

### DIFF
--- a/core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/PersonImp.java
+++ b/core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/PersonImp.java
@@ -145,13 +145,6 @@ public class PersonImp extends SingleVisualModelImp {
 
   @Override
   public void setSize(Dimension3 size) {
-    Dimension3 newScale = getScaleForSize(size);
-
-    if (newScale.hasNegativeComponents()) {
-      Logger.outln("Attempt to set person scale to " + newScale + " ignored.");
-      return;
-    }
-
-    this.setScale(newScale);
+    this.setScale(getScaleForSize(size));
   }
 }
--- a/core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/PersonImp.java
+++ b/core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/PersonImp.java
@@ -145,6 +145,13 @@ public class PersonImp extends SingleVisualModelImp {
 
   @Override
   public void setSize(Dimension3 size) {
-    this.setScale(getScaleForSize(size));
+    Dimension3 newScale = getScaleForSize(size);
+
+    if (newScale.hasNegativeComponents()) {
+      Logger.outln("Attempt to set person scale to " + newScale + " ignored.");
+      return;
+    }
+
+    this.setScale(newScale);
   }
 }
--- a/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Cylinder.java
+++ b/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Cylinder.java
@@ -245,27 +245,9 @@ public class Cylinder extends Shape {
     boundingSphere.radius = Math.sqrt(halfLengthSquared + maxRadiusSquared + maxRadiusSquared);
   }
 
-  public final BoundDoubleProperty length = new BoundDoubleProperty(this, 1.0) {
-    @Override
-    public void setValue(Double value) {
-      assert value >= 0.0 : value;
-      super.setValue(value);
-    }
-  };
-  public final BoundDoubleProperty bottomRadius = new BoundDoubleProperty(this, 1.0) {
-    @Override
-    public void setValue(Double value) {
-      assert value >= 0.0 : value;
-      super.setValue(value);
-    }
-  };
-  public final BoundDoubleProperty topRadius = new BoundDoubleProperty(this, 1.0) {
-    @Override
-    public void setValue(Double value) {
-      assert value >= 0.0 : value;
-      super.setValue(value);
-    }
-  };
+  public final BoundDoubleProperty length = new BoundDoubleProperty(this, 0.0);
+  public final BoundDoubleProperty bottomRadius = new BoundDoubleProperty(this, 0.0);
+  public final BoundDoubleProperty topRadius = new BoundDoubleProperty(this, 0.0);
   //todo: change default to CENTER?
   public final InstanceProperty<OriginAlignment> originAlignment = new InstanceProperty<OriginAlignment>(this, OriginAlignment.BOTTOM) {
     @Override
--- a/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Cylinder.java
+++ b/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Cylinder.java
@@ -245,9 +245,9 @@ public class Cylinder extends Shape {
     boundingSphere.radius = Math.sqrt(halfLengthSquared + maxRadiusSquared + maxRadiusSquared);
   }
 
-  public final BoundDoubleProperty length = new BoundDoubleProperty(this, 0.0);
-  public final BoundDoubleProperty bottomRadius = new BoundDoubleProperty(this, 0.0);
-  public final BoundDoubleProperty topRadius = new BoundDoubleProperty(this, 0.0);
+  public final BoundDoubleProperty length = new BoundDoubleProperty(this, 1.0);
+  public final BoundDoubleProperty bottomRadius = new BoundDoubleProperty(this, 1.0);
+  public final BoundDoubleProperty topRadius = new BoundDoubleProperty(this, 1.0);
   //todo: change default to CENTER?
   public final InstanceProperty<OriginAlignment> originAlignment = new InstanceProperty<OriginAlignment>(this, OriginAlignment.BOTTOM) {
     @Override
--- a/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Disc.java
+++ b/core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Disc.java
@@ -92,20 +92,8 @@ public class Disc extends Shape {
     }
 
   };
-  public final DoubleProperty innerRadius = new DoubleProperty(this, 0.0) {
-    @Override
-    public void setValue(Double value) {
-      assert value >= 0.0 : value;
-      super.setValue(value);
-    }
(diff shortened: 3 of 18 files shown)


dportnoy15:
The current changes won't completely prevent the bug yet, since they only prevent the radii (or similar in other shapes) from becoming negative if one of the size properties (width/height/depth) is modified. The radii can still be set to negative values by the user directly, triggering the assertion.

dabeshouse:
You point out this covers only part of the problem.
Do you have a plan for the other part and does it overlap with this at all?
In other words, should we review this PR now or wait?

dportnoy15:
I pushed another change that prevents the radii (and lengths) of various shapes from being set to negative directly, so the actual bug is now addressed, and the changes are ready for review.

core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/PersonImp.java
```diff
@@ -145,6 +145,13 @@ void updateNebPerson() {
 
   @Override
   public void setSize(Dimension3 size) {
-    this.setScale(getScaleForSize(size));
+    Dimension3 newScale = getScaleForSize(size);
+
+    if (newScale.hasNegativeComponents()) {
```
dabeshouse: Instead of doing this in each imp, can you catch it in `ModelSizePropertyController.updateAdapterFromUI()`?
Check the aimed for size and limit it to non-negative, or some tiny minimum.
Doing it there would mean both a single place to log this in the code and just one log entry.

core/story-api/src/main/java/org/lgna/story/implementation/AbstractCylinderImp.java
```diff
@@ -132,6 +138,11 @@ public Double getValue() {
 
     @Override
     protected void handleSetValue(Double value) {
+      if (value < 0.0) {
```
dabeshouse: I would like to find a similar answer to `setSize`, but am less sure. It might be possible to consolidate this logic in `SceneEditorUpdatingPropertyState` or `DoubleProperty` and reduce redundant code changes and extra logging, but I am not as convinced.
Take a look and see if there is something to be done for it.

dabeshouse (changes requested):
One recommendation. One place to explore further and maybe change.

core-nonfree/ide-nonfree/src/main/java/org/alice/stageide/personresource/PersonImp.java
dportnoy15: I looked at updateAdapterFromUI(), and doing the check there would probably make sense. If the aimed-for size is negative, should the operation just be cancelled (rather than shrinking the object to close-to-zero), or should the aimed-for size be modified to a spotive close-to-zero value instead (keeping the behavior currently in the PR)?

dabeshouse: I would say modify the target to some slightly greater than zero values. This way:
- The expected animation happens, to some extent at least.
- The proportions are maintained, while they would be lost at 0, 0, 0.

core/story-api/src/main/java/org/lgna/story/implementation/AbstractCylinderImp.java
dportnoy15: The only properties that use DoubleProperty (besides the ones I changed) are VrUser.scale and BubbleImp.portion, both of which sound like they should probably be limited to positive values. If there's a good chance that we'll want a DoubleProperty that can be negative at some point, I could create a child called PositiveDoubleProperty and implement a handleSetValue function with the negative check there.

jennej: I would at most leave this idea as a comment on DoubleProperty, Alice already has enough "just in case we need them" classes.

dabeshouse: Amen to not adding `AbstractJustInCaseClass`
If we want negative values we can perhaps add that as bounds options in a `DoubleProperty` constructor.
The common entry point seems to currently sit in `Property.animateValue()`. If we add (optional) bounds to `DoubleProperty` and override `animateValue` we can somewhat limit the needed changes.
Once the bounds are maintained on the `lgna.story` side we could drop the assorted anonymous subclasses of the (poorly named) `BoundedDoubleProperty` in `scenegraph`. They merely echo the values in `story`, with each anonymous subclass created to add an assert.

core/story-api/src/main/java/org/lgna/story/implementation/TorusImp.java
```diff
@@ -161,7 +161,7 @@ protected void handleSetValue(Double newInner) {
       }
     }
   };
-  public final DoubleProperty outerRadius = new DoubleProperty(TorusImp.this) {
+  public final DoubleProperty outerRadius = new DoubleProperty(TorusImp.this, DoubleProperty.CLOSE_TO_ZERO) {
```
dportnoy15, line 164: If this limit were set to 0 like the other cases, setting the property to 0 in the editor would still trigger an assertion, I think because it would cause the inner radius to become negative

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Cylinder.java
```diff
@@ -245,27 +245,9 @@ protected void updateBoundingSphere(edu.cmu.cs.dennisc.math.Sphere boundingSpher
...
-    }
-  };
-  public final BoundDoubleProperty topRadius = new BoundDoubleProperty(this, 1.0) {
-    @Override
-    public void setValue(Double value) {
-      assert value >= 0.0 : value;
-      super.setValue(value);
-    }
-  };
+  public final BoundDoubleProperty length = new BoundDoubleProperty(this, 0.0);
```
dabeshouse: This simplification is good to see. 

The limit was added to `org.lgna.story.implementation.DoubleProperty`, but this `BoundDoubleProperty` is a subclass of `edu.cmu.cs.dennisc.property.DoubleProperty` so the change from 1.0 to 0.0 is affecting the default value, not the limit.
We quickly replace that initial value, so we see no change in behavior.
That replacement vlaue comes from the `org.lgna.story.implementation.DoubleProperty` on `CylinderImp`, which does now have a lower limit.
Consider restoring the default values, which might make it less likely to be read as a lower limit.
A deeper fix would be to not have so many Property classes or so many layered related instances, but maybe not today.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Disc.java
```diff
@@ -92,20 +92,8 @@ public void setValue(Axis value) {
...
-    }
-  };
-  public final BoundDoubleProperty outerRadius = new BoundDoubleProperty(this, 1.0) {
-    @Override
-    public void setValue(Double value) {
-      assert value >= 0.0 : value;
-      super.setValue(value);
-    }
-  };
+  public final DoubleProperty innerRadius = new DoubleProperty(this, 0.0);
```
dabeshouse, line 95: The Cylinder comment applies here as well as on Sphere and Torus.

core/story-api/src/main/java/org/lgna/story/implementation/AbstractCylinderImp.java
```diff
@@ -124,7 +124,7 @@ public void setSize(Dimension3 size) {
 
   private final Cylinder sgCylinder = new Cylinder();
 
-  public final DoubleProperty length = new DoubleProperty(AbstractCylinderImp.this) {
+  public final DoubleProperty length = new DoubleProperty(AbstractCylinderImp.this, 0.) {
```
dabeshouse: The `0.` looks a little odd, why not `0` or `0.0`?

There are similar zero values in the other `*Imp` classes.

core/story-api/src/main/java/org/lgna/story/implementation/DoubleProperty.java
```diff
@@ -43,16 +43,41 @@
...
+import edu.cmu.cs.dennisc.java.util.logging.Logger;
+
 /**
  * @author Dennis Cosgrove
  */
 public abstract class DoubleProperty extends Property<Double> {
-  public DoubleProperty(PropertyOwnerImp owner) {
+  public static Double CLOSE_TO_ZERO = .01;
+
+  private Double minValue;
```
dabeshouse, line 55: Could be `final`

core/story-api/src/main/java/org/lgna/story/implementation/DoubleProperty.java
```diff
@@ -43,16 +43,41 @@
...
   protected Double interpolate(Double a, Double b, double portion) {
     return a + ((b - a) * portion);
   }
+
+  @Override
+  public void animateValue(final Double value, double duration, Style style) {
+    Double newValue = value;
+
+    if (!minValue.isNaN() && value < minValue) {
+      Logger.outln("Attempt to set double property below the minimum of " + minValue + " to " + value + " ignored.");
```
dabeshouse: I would be lying if I said I hate to be pedantic, so here is my pedantry.
The value is not ignored, because it will animate something, just stopping at a value that makes sense.
Maybe, `…+ minValue + " to " + value + " will stop at the minimum."`? Kind of awkward.
Or just ignore me, since this is only a minor bit of logging.

dabeshouse (commented):
Changes if you want to make them. Things to consider.

core/scenegraph/src/main/java/edu/cmu/cs/dennisc/scenegraph/Sphere.java
```diff
@@ -63,14 +63,5 @@ protected void updateBoundingSphere(edu.cmu.cs.dennisc.math.Sphere boundingSpher
     boundingSphere.radius = radius.getValue();
   }
 
-  public final BoundDoubleProperty radius = new BoundDoubleProperty(this, 0.5) {
-    @Override
-    public void setValue(Double value) {
-      if (value >= 0.0) {
-        super.setValue(value);
-      } else {
-        Logger.outln("Attempt to set sphere radius to " + value + " ignored.");
```
dabeshouse, line 72: CheckStyle
With this line gone the import for Logger should go too.

core/story-api/src/main/java/org/lgna/story/implementation/AbstractCylinderImp.java
dabeshouse: SphereImp and TorusImp still have `0.`

dabeshouse (approved):

PR #424, opened 2024-12-03
Feature/disable assertions

dabeshouse:
Disabled assertions in Alice builds.
Added a hack to push dialogs to the front.
This is long overdue, but should be tested well.


dportnoy15 (commented):
Does Alice have separate debug and release configs, and would it be hard to enable assertions for release, but not debug? Not sure if that's desirable, or if we want to move away from assertions.

dabeshouse:
> Does Alice have separate debug and release configs, and would it be hard to enable assertions for release, but not debug? Not sure if that's desirable, or if we want to move away from assertions.

We do not make builds that are debug, leaving that to individual developers as needed. I have left the `-ea` in the readme file.
The general guidance for release code is to have assertions turned off, so we will finally follow that in our official builds.

dportnoy15 (approved):

PR #426, opened 2025-02-20
Make different levels of nested menus behave the same way when the Escape key is pressed

dportnoy15:


core/ide/src/main/java/org/alice/ide/cascade/ExpressionCascadeManager.java
```diff
@@ -163,6 +163,10 @@ public ExpressionCascadeContext popContext() {
   }
 
   public ExpressionCascadeContext popAndCheckContext(ExpressionCascadeContext expectedContext) {
+    if (this.contextStack.size() == 0) {
```
dabeshouse, line 166: What is the case this is handling?
I can't seem to get in a state with an empty stack.

dabeshouse (approved):
Looks good. Works in my experiments.
Wondering about the empty context stack check, but it doesn't seem to be a problem.

PR #427, opened 2025-03-13
Feature/main menu standardization

dportnoy15:


core/croquet/src/main/java/org/lgna/croquet/views/imp/ScrollingPopupMenuUtilities.java
```diff
@@ -60,25 +63,53 @@ private ScrollingPopupMenuUtilities() {
...
+      if (menus.length > 0 && menus[menus.length - 1] instanceof JPopupMenu jPopupMenu) {
+        if (jPopupMenu.getLayout() instanceof ScrollingPopupMenuLayout scrollingPopupMenuLayout) {
+          int indexDelta = 0;
+          if (e.getKeyCode() == KeyEvent.VK_UP) {
+            indexDelta--;
+          }  else if (e.getKeyCode() == KeyEvent.VK_DOWN) {
+            indexDelta++;
+          }
+
+          scrollingPopupMenuLayout.adjustIndex(indexDelta);
```
dabeshouse: This can send a 0 and trigger a red queen when escape is pressed.
A conditional will help: `if (indexDelta != 0) {`

dabeshouse (changes requested):
In trying out the code I pressed escape when some menus were open. I pointed out one fix in a comment, but found the need for two more where a null UserActivity can lead to a Red Queen. These others are in `RtRoot.cancel` and `CascadeRoot.handleCancel`.
(Given your absence I may update the branch myself.)

core/croquet/src/main/java/org/lgna/croquet/views/imp/ScrollingPopupMenuUtilities.java
dportnoy15: I added checks for the user activity and the case where the popup menu is empty.

dabeshouse: No more red queen. :)

core/ide/src/main/java/org/alice/ide/recentprojects/RecentProjectsMenuModel.java
```diff
@@ -225,4 +225,12 @@ public void removeAllMenuItems() {
     }
     super.handleShowing(menuItemContainer, e);
   }
+
+  @Override
+  protected void initContents(MenuItemContainer menuItemContainer) {
```
dabeshouse, line 230: This is only happening once so it is not being updated as more worlds are opened.

core/croquet/src/main/java/org/lgna/croquet/SingleSelectListState.java
```diff
@@ -434,12 +429,6 @@ public List<T> createList() {
     return new List<T>(this);
   }
```
jennej, line 431: love deleted code

core/croquet/src/main/java/org/lgna/croquet/StaticMenuModel.java
```diff
@@ -54,9 +54,19 @@
  */
 public abstract class StaticMenuModel extends MenuModel {
   private StandardMenuItemPrepModel[] models;
+  private final boolean showScrollArrows;
 
   public StaticMenuModel(UUID individualId) {
```
jennej: this constructor version is only used the once, I'd just add the true to MethodHeaderMenuModel and drop this constructor option

core/ide/src/main/java/org/alice/ide/croquet/models/menubar/InternalTestingMenuModel.java
```diff
@@ -115,9 +115,31 @@ private static BooleanState createPickDebugFrameLazyIsFrameShowingState() {
...
-          BreakProjectAddNullMethodOperation.getInstance().getMenuItemPrepModel(), SEPARATOR, new CycleFrameSizeOperation().getMenuItemPrepModel(), SEPARATOR, ActiveTransactionHistoryComposite.getInstance().getIsFrameShowingState().getMenuItemPrepModel(),
+    super(UUID.fromString("6ee5bc6c-f45f-4eb9-bc4b-67fc524a05e8"),
+          false,
+          createSgDebugFrameLazyIsFrameShowingState().getMenuItemPrepModel(),
+          createGlrDebugFrameLazyIsFrameShowingState().getMenuItemPrepModel(),
+          createPickDebugFrameLazyIsFrameShowingState().getMenuItemPrepModel(),
+          SEPARATOR,
+          ShowMeHowToAddGalleryModelsIteratingOperation.getInstance().getMenuItemPrepModel(),
+          SEPARATOR, JavaCodeFrameComposite.getInstance().getIsFrameShowingState().getMenuItemPrepModel(),
```
jennej: add newlines after SEPARATOR to be consistent

core/croquet/src/main/java/org/lgna/croquet/views/imp/ScrollingPopupMenuUtilities.java
```diff
@@ -60,25 +63,50 @@ private ScrollingPopupMenuUtilities() {
...
+    @Override
+    public void menuKeyReleased(MenuKeyEvent e) {
+    }
+  };
+
   public static void initializeScrollingCapability(JPopupMenu jPopupMenu) {
     ScrollingPopupMenuLayout layout = new ScrollingPopupMenuLayout(jPopupMenu);
     jPopupMenu.setLayout(layout);
     jPopupMenu.addMouseWheelListener(mouseWheelListener);
-    jPopupMenu.add(new JScrollMenuItem(layout, ScrollDirection.UP), ScrollingPopupMenuLayout.ScrollConstraint.PAGE_START);
```
jennej, line 80: i think I prefer the more the more compact version that was there before?  

either way, remove the commented lines here

core/ide/src/main/java/org/alice/ide/recentprojects/RecentProjectsMenuModel.java
```diff
@@ -225,4 +225,12 @@ public void removeAllMenuItems() {
     }
     super.handleShowing(menuItemContainer, e);
   }
+
```
jennej, line 228: you didn't add a call to initContents here the same way you did in the other classes, is this intentional?

jennej (commented):
mostly just small comments.  I wish there was a better way to handle the duplicated
 menuItemContainer.getViewController().getAwtComponent().removeAll();
 menuItemContainer.getViewController().getAwtComponent().validate();
 but i suspect that is an evil i'm going to have to live with

core/croquet/src/main/java/org/lgna/croquet/views/imp/ScrollingPopupMenuUtilities.java
dportnoy15: The less compact code was only there when I thought maybe calling setFocusable was necessary, so I just reverted that change.

core/ide/src/main/java/org/alice/ide/recentprojects/RecentProjectsMenuModel.java
dportnoy15: In RecentProjectsMenuModel, both handleShowing and initContents call getChildren().

I could replace the call to setChildren() in RecentProjectsMenuModel.handleShowing() to initContents(), but the passed-in MenuItemContainer.getViewController() returns null, so initContents() would throw an exception. I could check for null, but then initContents would just be calling getChildren() anyway in that case.

If you think that would be a bit cleaner, I don't mind making the change though.

dportnoy15:
@jennej Regarding the duplicated removeAll() and validate() calls, I agree, but can't think of any great solutions. I did think of moving them into AbstractMenuModel, maybe into the initContents() there, and then having the subclasses call the super method first. However, then those two lines would then get called for all other subclasses of AbstractMenuModel too.

Just having a separate method, e.g. clearContents(), in AbstractMenuModel, and calling that from the relevant implementations of initContents() would work, but would make the code more difficult to read, and probably not worth it for just two lines.

core/croquet/src/main/java/org/lgna/croquet/PredeterminedMenuModel.java
```diff
@@ -53,11 +53,15 @@
 public abstract class PredeterminedMenuModel extends StaticMenuModel {
   private final StandardMenuItemPrepModel[] models;
 
-  public PredeterminedMenuModel(UUID individualId, StandardMenuItemPrepModel... models) {
-    super(individualId);
+  public PredeterminedMenuModel(UUID individualId, boolean showScrollArrows, StandardMenuItemPrepModel... models) {
```
dabeshouse: I am not a fan of booleans like the `showScrollArrows` boolean as part of the constructor. It does not change within  a class so it could be done as a checking method that is only overridden when they are to be shown.
Not sure I am asking for you to change this, since my distaste for the menu class hierarchy could lead me to suggest many more changes.
Something to consider before adding such flags in the future?

dabeshouse (changes requested):
You made me look at our menu code, but I suppose that is fair since I made you look at it first. So many opportunities for improvement here that I will not try to make as part of this PR.
Fix it so the recent projects list updates and consider any other suggestions.

dabeshouse:
lgtm
Jen, do you have anything feeling unresolved?

dabeshouse (approved):

core/ide/src/main/java/org/alice/ide/recentprojects/RecentProjectsMenuModel.java
jennej: ah, no, i just noticed that this one was different and wanted to make sure it was handled.  all good :)

PR #433, opened 2025-05-06
Feature/drop vector utils

dabeshouse:
Some code cleanup I avoided adding into the monster branch.


jennej (approved):
one less utility file! using generic angles that don't care how they're implemented! lovely

PR #439, opened 2025-06-06
Add some unit tests for Version.java

dportnoy15:


core/ast/src/test/java/org/lgna/project/VersionTest.java
```diff
@@ -0,0 +1,33 @@
...
+    public void testToString() {
+        String versionString = "3.14.0.4-alpha.2+build.local";
+        Version v = new Version(versionString);
+
+        assertEquals(versionString, v.toString());
+    }
+
+    @Test
+    public void testCurrentVersion() {
+        String currewntVersionString = TextFileUtilities.read(Version.class.getResourceAsStream("Version.txt")).trim();
```
dabeshouse: sp "current"

core/ast/src/test/java/org/lgna/project/VersionTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Test;
+
+import edu.cmu.cs.dennisc.java.io.TextFileUtilities;
+
+import static org.junit.Assert.*;
+
+public class VersionTest {
+
+    @Test
+    public void testIsValid() {
```
dabeshouse: Alice tests have (mostly) followed in the pattern of a single assert with a sometimes ridiculously long name that would describe the problem if it fails. This could be split into willReadValidVersionFromValidString and willReadInvalidVersionFomStringWithoutFourNumbers.

core/ast/src/test/java/org/lgna/project/VersionTest.java
```diff
@@ -0,0 +1,33 @@
...
+        assertTrue(new Version("3.14.0.4-alpha.2+build.local").isValid());
+        assertFalse(new Version("3vhs14.0.4-alpha.2+build.local").isValid());
```
dabeshouse, line 18: This bad value is clearly bad. Consider using "3.14.0.4alpha.2+build.local" either in place of in addition to.
It was my error to drop the - in front of the alpha that caused a bad version and triggered this discussion.

By the by, I tried making the same kind of error locally and your test below failed, so I would say it is a good thing

dabeshouse (approved):
This catches the kind of error I am concerned about me making again.
Consider my other comments, but generally looks good

dabeshouse (approved):

PR #441, opened 2025-06-06
add profile for building the installer

jennej:
only build installer with flag -DbuildInstaller=true

TheAliceProject/Planning#670


dabeshouse (approved):
Looks good
I used it to build on TeamCity and with the added profile it builds

PR #443, opened 2025-06-20
some small cleanups

jennej:


core/story-api/src/main/java/org/alice/interact/manipulator/ObjectTranslateDragManipulator.java
```diff
@@ -106,7 +106,7 @@ public void setOnscreenRenderTarget(OnscreenRenderTarget onscreenRenderTarget) {
 
   @Override
   public String getUndoRedoDescription() {
-    return "Object Move";
+    return "Obj Tranlate - Object Move";
```
dabeshouse: These make the undo stack clearer. Nice addition.
This one could use an `s`.

dabeshouse (changes requested):
One spelling change, but otherwise looks good to me.
(I expect I will make the edit)

dabeshouse (approved):

PR #448, opened 2025-07-14
code cleaning

dabeshouse:
Odds and ends that I did not want to drop in just before the release.


core/util/src/main/java/edu/cmu/cs/dennisc/java/io/FileUtilities.java
```diff
@@ -236,67 +146,13 @@ public static String getBaseName(File file) {
   }
 
   public static boolean exists(String path) {
```
dportnoy15: I'd suggest removing this method, both because writing FileUtilities.exists("...") takes more characters than new File("...").exists(), and because, according to IntelliJ, the method is unused anyway.

dabeshouse: Done

dportnoy15 (approved):

PR #449, opened 2025-07-29
When a world is opened, check if there are more recent backups availa…

dportnoy15:
…ble, and offer to load them instead

### DIFF
--- a/core/ide/src/main/java/org/alice/ide/ProjectApplication.java
+++ b/core/ide/src/main/java/org/alice/ide/ProjectApplication.java
@@ -370,9 +370,9 @@ public abstract class ProjectApplication extends PerspectiveApplication<ProjectD
 
   private final void loadProject(UserActivity activity, UriProjectLoader uriProjectLoader, boolean isLoadingBackups,
                                 Set<String> unloadableFiles) {
-    showWaitCursor();
     this.uriProjectLoader = uriProjectLoader;
     if (uriProjectLoader != null) {
+      showWaitCursor();
       uriProjectLoader.deliverContentOnEventDispatchThread(proj -> projectLoaded(activity, proj, isLoadingBackups,
               unloadableFiles));
     }
@@ -380,27 +380,29 @@ public abstract class ProjectApplication extends PerspectiveApplication<ProjectD
 
   private void projectLoaded(UserActivity activity, Project project, boolean isLoadingBackups,
                              Set<String> unloadableFiles) {
-    File saved = UriUtilities.getFile(getUri());
+    try {
+      File saved = UriUtilities.getFile(getUri());
 
-    boolean isBackup = false;
+      boolean isBackup = false;
 
-    if (!projectFileUtilities.isNewProject()) {
-      File parentDir = saved.getParentFile();
+      if (!projectFileUtilities.isNewProject()) {
+        File parentDir = saved.getParentFile();
 
-      if (parentDir != null) {
-        isBackup = BACKUP_EXTENSION.equals(getExtension(parentDir.getName()));
+        if (parentDir != null) {
+          isBackup = BACKUP_EXTENSION.equals(getExtension(parentDir.getName()));
+        }
       }
-    }
 
-    //File backupDir = projectFileUtilities.backupDirectory(saved, isBackup).toFile();
-    //boolean makeVrReady = uriProjectLoader.shouldMakeVrReady();
-
-    if (project == null) {
-      handleProjectLoadError(saved, activity, isBackup, isLoadingBackups, unloadableFiles);
-    } else {
-      handleProjectLoadSuccess(project, saved, activity, isBackup, isLoadingBackups, unloadableFiles);
+      if (project == null) {
+        handleProjectLoadError(saved, activity, isBackup, isLoadingBackups, unloadableFiles);
+      } else {
+        handleProjectLoadSuccess(project, saved, activity, isBackup, isLoadingBackups, unloadableFiles);
+      }
+    } catch (RuntimeException re) {
+      handleProjectLoadException(re, activity);
+    } finally {
+      hideWaitCursor();
     }
-    hideWaitCursor();
   }
 
   private void handleProjectLoadError(File projectFile, UserActivity activity, boolean isBackup,
@@ -474,23 +476,27 @@ public abstract class ProjectApplication extends PerspectiveApplication<ProjectD
         activity.cancel();
       }
     } catch (RuntimeException re) {
-      var message = new StringBuilder("Errors reported in " + getUri());
-      Throwable cause = re;
-      do {
-        var causeMessage = cause.getLocalizedMessage();
-        if (causeMessage != null) {
-          message.append("\n\n  ").append(causeMessage);
-        }
-        cause = cause.getCause();
-      } while (cause != null);
-      // TODO clear project remnants from system
-      uriProjectLoader = null;
-      activity.cancel(new CancelException(re));
-      Dialogs.showError("Unable to Load Project", message.toString());
-      setPerspective(getDocumentFrame().getNoProjectPerspective());
-      UserActivity newActivity = getOverallUserActivity().getLatestActivity().newChildActivity();
-      getDocumentFrame().getNewProjectOperation().fire(newActivity);
+      handleProjectLoadException(re, activity);
     }
+  }
+
+  private void handleProjectLoadException(RuntimeException re, UserActivity activity) {
+    var message = new StringBuilder("Errors reported in " + getUri());
+    Throwable cause = re;
+    do {
+      var causeMessage = cause.getLocalizedMessage();
+      if (causeMessage != null) {
+        message.append("\n\n  ").append(causeMessage);
+      }
+      cause = cause.getCause();
+    } while (cause != null);
+    // TODO clear project remnants from system
+    uriProjectLoader = null;
+    activity.cancel(new CancelException(re));
+    Dialogs.showError("Unable to Load Project", message.toString());
+    setPerspective(getDocumentFrame().getNoProjectPerspective());
+    UserActivity newActivity = getOverallUserActivity().getLatestActivity().newChildActivity();
+    getDocumentFrame().getNewProjectOperation().fire(newActivity);
   }
 
   private boolean createProjectFromBackup(File backup, File original) {
--- a/core/ide/src/main/java/org/alice/ide/ProjectApplication.java
+++ b/core/ide/src/main/java/org/alice/ide/ProjectApplication.java
@@ -44,10 +44,13 @@
 package org.alice.ide;
 
 import edu.cmu.cs.dennisc.java.awt.CursorUtilities;
+import edu.cmu.cs.dennisc.java.io.FileUtilities;
 import edu.cmu.cs.dennisc.java.lang.ClassUtilities;
 import edu.cmu.cs.dennisc.java.lang.SystemUtilities;
 import edu.cmu.cs.dennisc.java.net.UriUtilities;
 import edu.cmu.cs.dennisc.javax.swing.option.Dialogs;
+import edu.cmu.cs.dennisc.javax.swing.option.YesNoCancelResult;
+import org.alice.ide.croquet.models.projecturi.SaveAsProjectOperation;
 import org.alice.ide.frametitle.IdeFrameTitleGenerator;
 import org.alice.ide.project.ProjectDocumentState;
 import org.alice.ide.recentprojects.RecentProjectsListData;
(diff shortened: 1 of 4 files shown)


core/util/src/main/java/edu/cmu/cs/dennisc/java/io/FileUtilities.java
```diff
@@ -234,4 +238,14 @@ public static void copyFile(File in, File out) throws IOException {
...
     outChannel.close();
   }
+
+  public static LocalDateTime getModifiedDateTime(File f) {
+    try {
+      FileTime fileTime = Files.getLastModifiedTime(f.toPath());
+      return LocalDateTime.ofInstant(fileTime.toInstant(), ZoneId.systemDefault());
+    } catch(IOException ioe) {
+      ioe.printStackTrace();
+      return null;
```
dabeshouse: Returning null can produce an error.
Consider returning LocalDateTime.MIN.
Avoid NPEs and you can still check for it like you do null, but only if it matters.

core/ide/src/main/java/org/alice/ide/ProjectFileUtilities.java
```diff
@@ -33,6 +34,8 @@
 import static org.lgna.project.io.IoUtilities.PROJECT_EXTENSION;
 
 class ProjectFileUtilities {
+  public static final String PROJECT_FILE_EXTENSION = ".a3p";
```
dabeshouse: Could use `IoUtilities.PROJECT_EXTENSION` instead of (mostly) duplicating the string.

core/util/src/main/java/edu/cmu/cs/dennisc/java/io/FileUtilities.java
```diff
@@ -234,4 +238,14 @@ public static void copyFile(File in, File out) throws IOException {
     inChannel.close();
     outChannel.close();
   }
+
+  public static LocalDateTime getModifiedDateTime(File f) {
+    try {
+      FileTime fileTime = Files.getLastModifiedTime(f.toPath());
+      return LocalDateTime.ofInstant(fileTime.toInstant(), ZoneId.systemDefault());
+    } catch(IOException ioe) {
```
dabeshouse: Add a space after catch for style check.

core/ide/src/main/java/org/alice/ide/ProjectFileUtilities.java
```diff
@@ -159,12 +162,23 @@ private void backupActiveProject() throws IOException {
     removeExtraBackups(BACKUP_AUTO, backupDir);
   }
 
-  private Path backupDirectory(File saved) {
+  public boolean isNewProject() {
```
dabeshouse, line 165: The two callers of this are the projectApp that it calls back to for the URI and both callers negate the result.
Maybe this should be on ProjectApplication as `isNotNewProject()`

core/ide/src/main/java/org/alice/ide/ProjectFileUtilities.java
```diff
@@ -159,12 +162,23 @@ private void backupActiveProject() throws IOException {
...
+    URI uri = projectApp.getUri();
+
+    return uri != null && uri.getScheme().equalsIgnoreCase("gen");
```
dabeshouse: Use TemplateUriState.SCHEME for "gen".

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
   }
 
   public final void loadProject(UserActivity activity, UriProjectLoader uriProjectLoader) {
+    loadProject(activity, uriProjectLoader, false, new HashSet<>());
+  }
+
+  public final void loadProject(UserActivity activity, UriProjectLoader uriProjectLoader, boolean isLoadingBackups,
```
dabeshouse: Looks like this can be private

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -388,6 +487,83 @@ private void projectLoaded(UserActivity activity, Project project) {
...
+
+    return backup.toPath().getParent().resolveSibling(originalFileName).toFile();
+  }
+
+  private File getNextBackup(LocalDateTime modifiedTime, File backupDir, Set<String> unloadableFiles) {
+    if (backupDir == null) {
+      return null;
+    }
+
+    File[] backups = getSortedBackups("auto", backupDir);
```
dabeshouse: Could make `ProjectFileUtilities.BACKUP_AUTO` public and use it, also `getSortedBackups` is only called with "auto", does it need to be passed in?

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+              unloadableFiles));
     }
   }
 
-  private void projectLoaded(UserActivity activity, Project project) {
+  private void projectLoaded(UserActivity activity, Project project, boolean isLoadingBackups,
+                             Set<String> unloadableFiles) {
+    File saved = UriUtilities.getFile(getUri());
+
+    String parentDirName = "";
```
dabeshouse: Declare inside block where it is used.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+    boolean isBackup = false;
+
+    if (!projectFileUtilities.isNewProject()) {
+      parentDirName = saved.getParentFile().getName();
```
dabeshouse: Check for possible null from `getParentFile()`.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+      isBackup = BACKUP_EXTENSION.equals(getExtension(parentDirName));
+    }
+
+    File backupDir = null;
+
+    if (isBackup) {
+      backupDir = saved.getParentFile();
+    } else if (saved != null) {
```
dabeshouse: If saved can be null here then code further down will break.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+                  "WARNING: all backups more recent than the project were corrupted.\nWould you like to reload the original project file?")) {
+            loadProject(newProjectActivity(), new FileProjectLoader(mainProject, makeVrReady), false, unloadableFiles);
+          }
+        }
+      }
     } else {
+      // project was loaded successfully
+
+      if (isBackup && !isLoadingBackups) {
+        // User manually opened a backup, don't do anything special
```
dabeshouse: Why the empty true block?
Invert and fold in with line 432.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+
+    if (isBackup) {
+      backupDir = saved.getParentFile();
+    } else if (saved != null) {
+      backupDir = projectFileUtilities.backupDirectory(saved).toFile();
+    }
+
+    boolean makeVrReady = uriProjectLoader.shouldMakeVrReady();
+
     if (project == null) {
```
dabeshouse: This is a big method. Consider breaking out the true and false blocks into new methods to make it more readable.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+        // if it isn't empty, then we already attempted to load all newer backups,
+        // but weren't successful, so just continue loading the main project
+
+        LocalDateTime projectModifiedTime = FileUtilities.getModifiedDateTime(saved);
+
+        File backup = getNextBackup(projectModifiedTime, backupDir, unloadableFiles);
+
+        if (backup != null) {
+          // restart load with backup
+          if (Dialogs.confirmWithWarning("Load backup?",
```
dabeshouse: Collapse these two conditionals?

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,20 +363,108 @@ public UserActivity getOpenActivity() {
...
+          }
+        }
+      }
+
       try {
         updateInterface(project);
+
+        if (isLoadingBackups) {
+          // a backup was successfully loaded, prompt the user for what to do next
+          if (createProjectFromBackup(saved, getMainProjectFile(saved))) {
```
dabeshouse: Collapse these two ifs. You can move the comment outside and add, "If… a backup was…

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -388,6 +487,83 @@ private void projectLoaded(UserActivity activity, Project project) {
     hideWaitCursor();
   }
 
+  private boolean createProjectFromBackup(File backup, File original) {
+    YesNoCancelResult result = Dialogs.confirmOrCancel("Replace Project With Backup?",
+            "A backup has been opened successfully: " + backup.getName() + ".\n" +
```
dabeshouse: This needs to be tweaked to pass style check. Move +s to next line.
Also, all these strings should be internationalized, but that would be asking you to do as I say, not as I do.

dabeshouse (changes requested):
Lots of little fiddly suggestions.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
dportnoy15: I'll do some more testing to see what happens in different situations and add some more checks.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
dportnoy15: I add some code to that block in uncommitted changes, but I can fold it in for now, if you prefer

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -388,6 +487,83 @@ private void projectLoaded(UserActivity activity, Project project) {
...
+                    "Would like to replace the original project, or create a new project from the backup?\n" + "" +
+                    "Cancel to do neither and just continue opening the backup.");
+
+    if (result == YesNoCancelResult.YES || result == YesNoCancelResult.NO) {
```
dabeshouse: This is the same as != CANCEL - as the comment on the else block below says.
Consider using a switch instead.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
dportnoy15: Since I should probably check for non-auto backups as well, which I just hadn't looked at yet, I'd like to leave the parameter there for now. I suspect that the function should probably just check all backups anyway, so I may end taking the check out of the function completely anyway.

core/ide/src/main/java/org/alice/ide/ProjectFileUtilities.java
dportnoy15: I would like to, at least, postpone this change. In code that I will add later, I do use the non-negated version. I would like to decide whether to negate the function once all the code is in place.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -352,40 +365,210 @@ public UserActivity getOpenActivity() {
...
-          var causeMessage = cause.getLocalizedMessage();
-          if (causeMessage != null) {
-            message.append("\n\n  ").append(causeMessage);
-          }
-          cause = cause.getCause();
-        } while (cause != null);
-        // TODO clear project remnants from system
+      handleProjectLoadSuccess(project, saved, activity, isBackup, isLoadingBackups, unloadableFiles);
+    }
+    hideWaitCursor();
```
dabeshouse: Having this here means it only happens if `uriProjectLoader` is not null (line 375)
The show and hide should be more tightly coupled, perhaps with the hide in a finally block to make sure it happens.

dabeshouse:
Before this work started the responsibilities involved here were not well contained and not what I would hold up as good OO.
That has not made your work any simpler.
I look at the calling back and forth between `ProjectApplication`, `ProjectFileUtilities`, `FileUtilities`, and the `UriProjectLoader` hierarchy, especially `FileProjectLoader` and want to split it into more meaningful responsibilities.

This comment is not going anywhere beyond an apology for the state of things.
I am not requesting you do this, but if you are inclined you have my support.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
dabeshouse: Adding the finally block does ensure its execution, but does not address the coupling of show/hide.
You could:
- Move the `finally … hide` to the calling above method with show.
- Move the show into here (but this is already pretty big.
- Create a unified method to wrap the show and hide, maybe `showWaitCursorWhileDoing(Runnable r)`

dabeshouse (approved):

PR #450, opened 2025-07-30
another pass removal PR

jennej:


alice-ide/src/main/java/org/alice/stageide/EntryPoint.java
```diff
@@ -130,15 +130,11 @@ public static void main(final String[] args) {
...
-        if ("null".equalsIgnoreCase(args[0])) {
-          //pass
-        } else {
-          if ("-l".equalsIgnoreCase(args[0])) {
-            index = 1;
-            if (args.length > 1) {
-              localeString = args[1];
-              index = 2;
-            }
+        if (!"null".equalsIgnoreCase(args[0]) && "-l".equalsIgnoreCase(args[0])) {
```
dportnoy15: Isn't the comparison to "null" unnecessary here since if it equals "-l", it obviously doesn't equal "null"?

core/ast/src/main/java/org/lgna/project/ast/AstUtilities.java
```diff
@@ -150,21 +150,16 @@ public static Iterable<JavaMethod> getKeyMethods(ArgumentListProperty<JavaKeyedA
   }
 
   public static boolean isKeyedArgumentListPropertyComplete(ArgumentListProperty<JavaKeyedArgument> argumentListProperty) {
+    boolean isFound = false;
```
dabeshouse: This changes the logic.
Previously it would check each method and if any were not found it would return false.
Now if any are found it returns true.

This applies when a user is adding optional arguments to a method call.
For example, if you set the font on a say, you can no longer set the duration.

This method might make more sense if `getKeyMethods` was renamed something like `getOptionalParameters`. Yes the objects are each a `JavaMethod`, but they are methods that represent a parameter, like `DurationAnimationStyleAsSeenByArgumentFactory.asSeenBy`

dabeshouse (changes requested):
One bug to fix, but otherwise looks good.

alice-ide/src/main/java/org/alice/stageide/EntryPoint.java
jennej: this whole function was mildly terrifying so I didn't read it properly, but you're right

dabeshouse (approved):
lgtm

PR #454, opened 2025-08-14
When checking for a new project, include both blank and starter worlds

dportnoy15:


core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -409,7 +410,6 @@ private void projectLoaded(UserActivity activity, Project project, boolean isLoa
   private void handleProjectLoadError(File projectFile, UserActivity activity, boolean isBackup,
                                       boolean isLoadingBackups, Set<String> unloadableFiles) {
     File backupDir = projectFileUtilities.backupDirectory(projectFile, isBackup).toFile();
-    boolean makeVrReady = uriProjectLoader.shouldMakeVrReady();
```
dabeshouse: You moved this down to where it is used, but the next line sets uriProjectLoader to null. Am I missing something or will that be a problem?

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -462,7 +461,7 @@ private void handleProjectLoadSuccess(Project project, File projectFile, UserAct
         activity.cancel();
 
         // restart load with backup
-        loadProject(newProjectActivity(), new FileProjectLoader(backup, makeVrReady), true, unloadableFiles);
+        loadProject(newProjectActivity(), new FileProjectLoader(backup, uriProjectLoader.shouldMakeVrReady()), true, unloadableFiles);
```
dabeshouse: Similar to above I am concerned about setting uriProjectLoader to null a few lines above this.

dabeshouse (commented):
Setting `uriProjectLoader` to null was a problem when I ran this. Is there more on your next branch that addresses this?

dabeshouse (approved):

PR #457, opened 2025-08-22
Improve how world backups are made and used

dportnoy15:
- Create automatic backups for new, unsaved projects
 - Only create automatic backups when there have been changes since the last save
 - If a project fails to load, try loading its backups until one is successfully loaded


core/ide/src/main/java/org/alice/ide/uricontent/FileProjectLoader.java
```diff
@@ -65,6 +65,10 @@ public URI getUri() {
     return getSaveFile().toURI();
   }
 
+  @Override
+  protected void handleLoadException(File file, Exception e) {
```
dabeshouse, line 69: Maybe add a comment about the logic of ignoring this exception?

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -386,75 +385,101 @@ private final void loadProject(UserActivity activity, UriProjectLoader uriProjec
...
-    if (!uriProjectLoader.isNewProject()) {
-      File parentDir = saved.getParentFile();
+    if (backup != null) {
+      boolean loadBackup = Dialogs.confirmWithWarning("Load Backup of New Project",
+              "WARNING: Backups of an unsaved new project were detected. Would you like to load the latest one?");
 
-      if (parentDir != null) {
-        isBackup = BACKUP_EXTENSION.equals(getExtension(parentDir.getName()));
+      if (loadBackup) {
+        loadProject(newProjectActivity(), new FileProjectLoader(backup, false), true, true, new HashSet<>());
```
dabeshouse: Using this `loadProject` leads to the user working with, and seeing the full backup directory (e.g. `/Users/dabeshou/Alice3/MyProjects/.defaultbak/auto20250825_145235.a3p`)
File systems can be difficult enough for beginners, and hidden directories even more so.

It would be better if they were restored to the state of having an unsaved project without a file name.
Another alternative would be to require the user to immediately save the backup and have the dialog prefilled with something like `…/Alice3/MyProjects/RecoveredWorld.a3p`

Also, when I tried the code as is and saved the restored world I ended up with an extra layer of backups named for the default autosave:
- MyProjects/NewSave.a3p
- MyProjects/NewSave.bak
- MyProjects/NewSave.bak/auto20250825_145235.a3p
- MyProjects/NewSave.bak/auto20250825_145235.bak/save20250826_091120.a3p

core/ide/src/main/java/org/alice/ide/ProjectDocumentFrame.java
```diff
@@ -261,7 +261,7 @@ public Operation getResourcesDialogLaunchOperation() {
   }
 
   public BooleanState getStatisticsFrameIsShowingState() {
-    return this.stasticsFrameIsShowingState;
+    return this.statisticsFrameIsShowingState;
```
dabeshouse, line 264: Spelling fixes are always a good thing.
(OK, maybe not always if they are around serialization, but we won't think about that)
(Or reflection, fixes can be bad there too)
But this is a good fix, thank you.

core/ide/src/main/resources/org/alice/ide/croquet/models/projecturi/croquet.properties
```diff
@@ -1,3 +1,22 @@
+BackupProjectOperation.BackupLoadErrorAndLoadOriginalDialog.message = WARNING: all backups more recent than the project were corrupted.\nWould you like to reload the original project file (</projectName/>)?
```
dabeshouse: With these strings gathered it is easier to see inconsistencies.
So I suggest making this, and UnsavedBackupOpenedDialog.message start with upper case letters.
Also, end sentences with a period.
Or not, but try to stay consistent

core/ide/src/main/java/org/alice/ide/frametitle/AliceIdeFrameTitleGenerator.java
```diff
@@ -53,18 +54,30 @@
...
+
+    URI uri = null;
+    URI mainProjectUri = null;
+
+    if (projectLoader != null) {
+      uri = projectLoader.getUri();
+      mainProjectUri = projectLoader.getMainProjectUri();
+    }
+
+    if (mainProjectUri != null) {
```
dabeshouse: This will only not be null if line 69 executes and returns something.
That means you could move this block inside there, and probably eliminate the uri variable.

core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
```diff
@@ -0,0 +1,84 @@
...
+
+        Dialogs.showError(title, message);
+    }
+
+    public boolean showProjectLoadErrorAndLoadBackupDialog(String projectName, boolean isCurrentProjectBackup) {
+        String title = findLocalizedText("ProjectLoadErrorAndLoadBackupDialog.title");
+        String message = findLocalizedText("ProjectLoadErrorAndLoadBackupDialog.message")
+                .replaceAll("</projectName/>", projectName)
+                .replaceAll("</backupType/>", isCurrentProjectBackup
+                        ? "an earlier backup"
```
dabeshouse, line 60: The backupType is not localized.
You could add two more entries, but a different language might order things differently and this would break down.
Better you could make two at the message level, ProjectLoadErrorAndLoadBackupDialog.messageCurrent and ProjectLoadErrorAndLoadBackupDialog.messageEarlier.

core/ide/src/main/java/org/alice/ide/uricontent/UriProjectLoader.java
```diff
@@ -89,5 +98,59 @@ public boolean shouldMakeVrReady() {
...
+      return f;
+    }
+
+    File backupDir = f.getParentFile();
+    String originalFileName = FileUtilities.getBaseName(backupDir) + "." + PROJECT_EXTENSION;
+
+    return f.toPath().getParent().resolveSibling(originalFileName).toFile();
+  }
+
+  protected String getParentDirExtension(File f) {
```
dabeshouse: Could be private

core/ide/src/main/java/org/alice/ide/uricontent/UriProjectLoader.java
```diff
@@ -89,5 +98,59 @@ public boolean shouldMakeVrReady() {
...
+    if (f == null) {
+      return false;
+    }
+
+    String parentDirExtension = getParentDirExtension(f);
+
+    return DEFAULT_BACKUP_DIR.equals(parentDirExtension);
+  }
+
+  protected File getMainProjectFile(File f) {
```
dabeshouse: Could be private

core/ide/src/main/java/org/alice/ide/uricontent/UriProjectLoader.java
```diff
@@ -89,5 +98,59 @@ public boolean shouldMakeVrReady() {
...
+    File projectFile = UriUtilities.getFile(getUri());
+
+    if (projectFile != null) {
+      uri = getMainProjectFile(projectFile).toURI();
+    }
+
+    return uri;
+  }
+
+  public boolean isBackup(File f) {
```
dabeshouse: I think the file passed in (here and other methods) is always the file associated with this loader. Usually fetched by `UriUtilities.getFile(projectApp.getUri())`.
If it is not the file for this loader, this will probably give the wrong result since getParentDirExtension checks if this loader is a new project.
It may also be the case that this is only called on the subclass, FileProjectLoader.

Not sure what to do with this now that isn't too large, especially since this is due to pre-existing conditions.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -198,7 +198,7 @@ public void handleVersionNotSupported(File file, VersionNotSupportedException vn
     Dialogs.showUnableToOpenFileDialog(file, String.format("%s is not backwards compatible with:\n    File Version: %s\n    (Minimum Supported Version: %s)", getApplicationName(), vnse.getVersion(), vnse.getMinimumSupportedVersion()));
   }
 
-  private UriProjectLoader uriProjectLoader;
+  public UriProjectLoader uriProjectLoader;
```
dabeshouse: Not a good thing to expose internals like this.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -245,6 +246,15 @@ protected boolean isProjectUpToDateWithSceneSetUp() {
     }
   }
 
+  protected boolean isProjectUpToDateWithBackups() {
+    UndoHistory history = this.getProjectHistory();
+    if (history == null) {
```
dabeshouse: Simplifies to `return history == null || this.projectHistoryIndexBackups == history.getInsertionIndex()`
This also looks a lot like the two methods above. Could extract `isProjectUpToDateWith(int index)`.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -265,6 +275,13 @@ protected void updateHistoryIndexSceneSetUpSync() {
     }
   }
 
+  private void updateHistoryIndexBackupSync() {
```
dabeshouse, line 270: Like the up to date check above this adds a third form. Consider extracting int getHistoryIndex() for use in this and the two update methods above.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -503,65 +558,69 @@ private void handleProjectLoadException(RuntimeException re, UserActivity activi
...
         } catch (IOException ioe) {
           Dialogs.showError("Unable to save file", ioe.getMessage());
         }
+      }
+      case NO -> {
+        // when the user saves it, create a new project from this one
 
-        yield true;
+        try {
+          uriProjectLoader = new StarterProjectFileLoader(new URI("starterfile:/"), uriProjectLoader.shouldMakeVrReady());
```
dabeshouse: After doing this a call to `updateTitle()` is probably in order.

dabeshouse (commented):
Mostly done.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
dabeshouse: Been thinking more about this.
I am not sure starter is the way to go. This is a backup of a file project, not necessarily a starter and using starter will probably complicate logic in other places.
I am not sure what an isolated fix is, perhaps we should discuss more.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -386,100 +390,141 @@ private final void loadProject(UserActivity activity, UriProjectLoader uriProjec
...
+
+      return;
+    }
+
+    if (backup != null) {
+      if (backupProjectOperation.showProjectLoadErrorAndLoadBackupDialog(projectFile.getName(), isBackup)) {
+        loadProject(newProjectActivity(), new FileProjectLoader(backup, makeVrReady), true, isMainProjectCorrupted, unloadableFiles);
+      }
+    } else {
+      if (isMainProjectCorrupted) {
```
dabeshouse, line 481: If you get to this point you have tried all the backups and are notifying the user there is nothing to do.
After that we should return to the open file dialog as done at the end of `handleProjectLoadException`
You could extract the last three lines there into a common method to use here and return to the open dialog.

core/ide/src/main/java/org/alice/ide/frametitle/AliceIdeFrameTitleGenerator.java
```diff
@@ -42,31 +42,32 @@
...
+    if (projectLoader != null) {
+      isBackup = projectLoader.isBackup();
+      File file = projectLoader.getMainProjectFile();
       if (file != null) {
         sb.append(file);
       }
-      sb.append(" ");
     }
-    if (!isDocumentUpToDateWithUri) {
+    if (isBackup || !isDocumentUpToDateWithUri) {
```
dabeshouse, line 70: This change does cause the * to show up, but it removes consistent meaning from it because it does not have any effect as far as closing the world.
If it is a backup and you have not saved it it will not stop you, which is not what the * generally implies.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -386,100 +390,136 @@ private final void loadProject(UserActivity activity, UriProjectLoader uriProjec
...
-      File mainProject = getMainProjectFile(projectFile);
+    // If this failed attempt was already a load of a backup, keep the existing value
+    // Otherwise, this is the main project, so set this to true
+    if (!isLoadingBackups) {
+      isMainProjectCorrupted = true;
+    }
 
-      LocalDateTime projectModifiedTime = FileUtilities.getModifiedDateTime(mainProject);
+    File mainProject = uriProjectLoader.getMainProjectFile();
+    LocalDateTime projectModifiedTime = FileUtilities.getModifiedDateTime(mainProject);
```
dabeshouse, line 461: Something changed the way this works.
The value of mainProject may be null, if the backup was not loaded, which causes getModifiedTime to throw and an uphelpful error dialog to show and then fall through to the main open window.
Previously there was no error and the user was given the chance to open earlier backups.
 
I think it was in the latest commit "Always show an asterisk in the title bar when a backup is open" 3d5c890e24c0bef2d05e4392ba7827bf5e869eb6

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -51,12 +51,14 @@
 import edu.cmu.cs.dennisc.java.util.logging.Logger;
 import edu.cmu.cs.dennisc.javax.swing.option.Dialogs;
 import edu.cmu.cs.dennisc.javax.swing.option.YesNoCancelResult;
-import org.alice.ide.croquet.models.projecturi.SaveAsProjectOperation;
+import org.alice.ide.croquet.models.projecturi.BackupProjectOperation;
 import org.alice.ide.frametitle.IdeFrameTitleGenerator;
 import org.alice.ide.project.ProjectDocumentState;
 import org.alice.ide.recentprojects.RecentProjectsListData;
 import org.alice.ide.uricontent.FileProjectLoader;
+import org.alice.ide.uricontent.StarterProjectFileLoader;
```
dabeshouse: Stylecheck - unused

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -80,19 +82,18 @@
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
```
dabeshouse: Stylecheck - unused

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -607,17 +644,32 @@ private void updateInterface(Project project) {
 
     updateHistoryIndexFileSync();
     updateUndoRedoEnabled();
-    projectFileUtilities.startAutoSaving();
+
+    if (!uriProjectLoader.isBackup()) {
```
dabeshouse: Skipping automatic backups of backups makes sense, but this check also skips the default backups before a project is saved.
Since you have added a save on exit this is less of a concern, but that save is not enough if Alice freezes or crashes.

dabeshouse (approved):
lgtm
Merging

PR #460, opened 2025-08-27
Fix binary operation LHS bug

dabeshouse:
The keyboard navigation fix to clear menus properly when the escape key is pressed caused problems when building binary operations (AND, OR, plus, etc.) and possibly other menu results.
This changes the esc notification from making the menu invisible directly, to canceling the UserActivity.

This has two effects on the UI.
- First, fixing the bug, since the RtRoot will no longer look for mere invisibility and instead look for the activity. This also means the NULL_CONTEXT work around can be dropped from one place.
- Second, escape will cancel only a single menu most of the time. This might be useful in navigation, or it might be annoying for users. I think I like it, but please express an opinion when reviewing this.


dportnoy15 (approved):
I also prefer when the escape key closes only one level of a menu

PR #466, opened 2025-09-25
remove the sticky note ui/ cleanup how we draw the highlights for the find function

jennej:
found a herd of yaks, deleted a lot of code


dabeshouse (approved):
More unused code removed. Thank you.
lgtm

PR #467, opened 2025-09-26
Accept null expressions. Allows reading in of worlds with certain bugs.

dabeshouse:
This allows reading of worlds with null expressions.
The recent LHS bug is one way to create and write worlds with this issue.


jennej (approved):
well, that was easy

PR #468, opened 2025-09-26
Save the normalized vectors instead of creating them and using originals

dabeshouse:
An error I introduced with the immutable math change.
These vectors are used to create orthogonal matrices, which are in turn not normalized.
So far it has led to no errors more severe than logging that the matrices are not normalized.


jennej (approved):
lgtm

PR #469, opened 2025-09-26
Hotfix/project read

dabeshouse:
Two fixes to allow programs with errors to be read in.
Adding to hotfix release that has a correction to keep from creating such errors.

### DIFF
--- a/core/ast/src/main/java/org/lgna/project/virtualmachine/VirtualMachine.java
+++ b/core/ast/src/main/java/org/lgna/project/virtualmachine/VirtualMachine.java
@@ -138,20 +138,24 @@ public abstract class VirtualMachine {
     return getConstructor(entryPointType, arguments).evaluate(this, null, arguments);
   }
 
-  public Object createAndSetFieldInstance(UserInstance userInstance, UserField field) {
-    Expression expression = field.initializer.getValue();
-    assert expression != null;
-    Object rv = this.evaluate(expression);
-    userInstance.setFieldValue(field, rv);
-    return rv;
+  public void createAndSetFieldInstance(UserInstance userInstance, UserField field) {
+    try {
+      Object value = evaluate(field.initializer.getValue());
+      userInstance.setFieldValue(field, value);
+    } catch (RuntimeException e) {
+      if (isForRunning) {
+        throw e;
+      }
+      Logger.warning("Error when setting up scene " + e.getMessage());
+    }
   }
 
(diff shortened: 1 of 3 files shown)


core/ide/src/main/java/org/alice/ide/IDE.java
```diff
@@ -485,6 +485,7 @@ protected VirtualMachine createVirtualMachineForSceneEditor() {
 
   public final VirtualMachine createRegisteredVirtualMachineForSceneEditor() {
     VirtualMachine vm = this.createVirtualMachineForSceneEditor();
```
jennej, line 487: i feel like the function name createVirtualMachineForSceneEditor would imply that i've created a vm for the scene editor, and i shouldn't need to call a second function to create it even harder for the scene editor?

core/ast/src/main/java/org/lgna/project/virtualmachine/VirtualMachine.java
```diff
@@ -1237,6 +1241,15 @@ public List<VirtualMachineListener> getVirtualMachineListeners() {
...
+    isForRunning = false;
+  }
+
   private final List<VirtualMachineListener> virtualMachineListeners = Lists.newLinkedList();
   private boolean isStopped = false;
+
+  // Marks this VM for use in running worlds. When true it allows errors to be thrown that interrupt execution.
+  // A value of false indicates this VM is used during scene loading or scene setup where thrown exceptions can
+  // cause these processes to break and should be simply logged and the setup code continued.
+  private boolean isForRunning = true;
```
jennej, line 1254: nitpicky name question, should this variable maybe be named something like "allowExceptions" or "quietlyLogErrors"

jennej (commented):
i have no issues with the behavior, i'm just overthinking the names

core/ide/src/main/java/org/alice/ide/IDE.java
dabeshouse: But that one line `createVirtualMachineForSceneEditor` method can be used in other places!
It's not, but it could be.
And it's `protected` so we can override it.
We don't, but we could.

core/ast/src/main/java/org/lgna/project/virtualmachine/VirtualMachine.java
dabeshouse: I went with intent over effect. There may be other behavior changes that a VM at runtime could imply where checking something about error handling would make less sense.
Either way, since this is me cherry picking a commit from develop I will keep it as is so future merges will be less painful.

PR #470, opened 2025-10-03
Drop unused code paths and don't log lack of drop site.

dabeshouse:
Removing some dead code and distracting logging.
Drops in the scene have a dropsite, but code drops do not so logging this every time is meaningless.


jennej (approved):
lgtm

PR #473, opened 2025-10-10
Show localized timestamps instead of filenames for project backups

dportnoy15:


core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
```diff
@@ -54,11 +71,23 @@ public void showBackupLoadErrorDialog() {
...
+
+        if (backupDateString == null) {
+            message = findLocalizedText("ProjectLoadErrorAndLoadBackupDialog.message")
+                    .replaceAll("</projectName/>", projectName);
+        }   else {
+            message = findLocalizedText("ProjectLoadErrorAndLoadBackupDialog.messageWithDate")
+                    .replaceAll("</backupDate/>", backupDateString);
+        }
+
+        message = message.replaceAll("</backupType/>", isCurrentProjectBackup
```
dabeshouse: This snippet is unlocalized.

Also I am not sure how much sense this distinction will convey in the new `messageWithDate`:
> An automated backup from Oct 3 could not be loaded.
> Would you like to try loading a backup?

Thirdly, why "automated"?
Are we are distinguishing between those made against user saves versus the intermittent ones? It doesn't look that way.
Is this just all backups, since they are all automated?

core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
```diff
@@ -92,4 +121,19 @@ public boolean showMoreRecentBackupsDialog() {
...
         return Dialogs.confirmWithWarning(title, message);
     }
+
+    private String getDateStringFromBackupName(String name) {
+        if (name.length() < 8) {
+            return null;
+        }
+
+        try {
+            String datetime = name.substring(4, name.length() - 4);
```
dabeshouse, line 134: There is also the single argument version of `substring` that just needs the `beginIndex`.

core/ide/src/main/java/org/alice/ide/ProjectFileUtilities.java
```diff
@@ -36,9 +36,9 @@ public class ProjectFileUtilities {
   public static final String BACKUP_AUTO = "auto";
   public static final String BACKUP_EXTENSION = "bak";
   public static final String DEFAULT_BACKUP_DIR = "defaultbak";
+  public static final DateTimeFormatter ORDER_FORMAT = DateTimeFormatter.ofPattern("yyyyMMdd_HHmmss");
```
dabeshouse, line 39: Did you mean to move this line? No need to move it back, just curious.

core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
```diff
@@ -54,11 +71,23 @@ public void showBackupLoadErrorDialog() {
...
-                        ? "an earlier backup"
-                        : "a backup");
+        String message = "";
+
+        String backupDateString = isCurrentProjectBackup
+                ? getDateStringFromBackupName(projectName)
+                : null;
+
+        if (backupDateString == null) {
+            message = findLocalizedText("ProjectLoadErrorAndLoadBackupDialog.message")
```
dabeshouse, line 91: What are the conditions to end up here?
Do we have backups without valid dates that we do try to read?

dabeshouse (commented):
Looks good.
A few pieces to discuss.

core/ide/src/main/java/org/alice/ide/ProjectFileUtilities.java
dportnoy15: Yes, I moved it so it would be grouped with the public constants

core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
dportnoy15: To my knowledge, we don't have backups with invalid dates, but getDateStringFromBackupName() could conceivably return null. Also, in this particular case, backupDateString will be null if the project isn't a backup.

core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
dportnoy15: The second argument here is the end index, not the length. I'm stopping 4 chars before the end, to not include .a3p. It just so happens that I'm skipping both the first 4 and the last 4 characters.

core/ide/src/main/java/org/alice/ide/croquet/models/projecturi/BackupProjectOperation.java
dportnoy15: Is the example you provided something you actually saw? As far as I cna tell, it should be impossible to get that since, if the project is a backup, it should show the human-readable date and say "an earlier backup".

I can localize that snippet, although a better solution would be to have each possible full message localized, since concatenating localized snippets that work in English will probably have words out of order in other languages.

Right now, there are four possible messages this dialog could show (paraphrased for brevity):
1. Project failed to load. Do you want to load a backup?
2. Project failed to load. Do you want to load an earlier backup?
3. Backup from date failed to load. Do you want to load a backup?
4. Backup from date failed to load. Do you want to load an earlier backup?

Option 3 should never actually happen (unless you did actually somehow see it), and option 2 should only happen if there's an error parsing the date from the filename. Given all of that, I could make 3 localized messages instead of the current two, and avoid doing a replace for the (currently unlocalized) snippet.

In that case, if a backup fails to load, the message would always end with "Do you want to load an earlier backup?". The reason I originally had that phrasing was to distinguish between the "main project" failing to load, and a backup failing to load (and then the system trying to load *a different* backup), and to clarify the difference when one message is shown right after the other.

I don't feel particularly strongly about that though, so we could just always say "a backup".

Regarding the "automated backup" bit, I just took that from the example in the [ticket](https://github.com/orgs/TheAliceProject/projects/5/views/15?filterQuery=milestone%3A%22Mock+Turtle%22+assignee%3Adportnoy15&pane=issue&itemId=132991431&issue=TheAliceProject%7CPlanning%7C816), but I hav eno objection to just saying "backup" and I agree that the current phrasing suggests other types of backups.

dabeshouse (approved):
lgtm

PR #480, opened 2025-10-23
When a program is shut down, set the active scene to null to facilita…

dportnoy15:
…te the release of event handlers


dabeshouse (approved):
lgtm

PR #486, opened 2025-10-28
even more code cleanup

jennej:


dabeshouse (changes requested):
The changes look good, but there are a few references to javamp3 left.
- The root pom.xml
- The story-api pom.xml
- In netbeans pom.xml and Alice3Library.xml

dabeshouse (approved):
lgtm

PR #487, opened 2025-10-30
Ensure garbage collection after an Alice world stops running

dportnoy15:
Make sure all relevant objects are garbage-collected after an Alice world is closed in the editor

### DIFF
--- a/core/croquet/src/main/java/org/lgna/croquet/views/AwtContainerView.java
+++ b/core/croquet/src/main/java/org/lgna/croquet/views/AwtContainerView.java
@@ -102,10 +102,6 @@ public abstract class AwtContainerView<J extends Container> extends AwtComponent
     return rv;
   }
 
-  public int getComponentZOrder(AwtComponentView<?> component) {
-    return this.getAwtComponent().getComponentZOrder(component.getAwtComponent());
-  }
-
   public void setComponentZOrder(AwtComponentView<?> component, int index) {
     this.checkEventDispatchThread();
     this.getAwtComponent().setComponentZOrder(component.getAwtComponent(), index);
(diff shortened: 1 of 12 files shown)


core/croquet/src/main/java/org/lgna/croquet/views/AwtContainerView.java
```diff
@@ -102,10 +102,6 @@ public AwtComponentView<?>[] getComponents() {
     return rv;
   }
 
-  public int getComponentZOrder(AwtComponentView<?> component) {
-    return this.getAwtComponent().getComponentZOrder(component.getAwtComponent());
-  }
-
   public void setComponentZOrder(AwtComponentView<?> component, int index) {
```
dabeshouse: You took out the unused getter. The setter can go too.

core/ide/src/main/java/org/alice/stageide/program/RunProgramContext.java
```diff
@@ -71,7 +71,9 @@ public Container getContainer() {
 
   @Override
   public void cleanUpProgram() {
-    this.getContainer().removeAll();
+    if (this.getContainer() != null) {
+      this.getContainer().removeAll();
```
dabeshouse: Will this ever get called now?
Is the container always null?

core/util/src/main/java/edu/cmu/cs/dennisc/animation/AbstractAnimator.java
```diff
@@ -65,6 +65,7 @@ public abstract class AbstractAnimator implements Animator {
   private double tCurrent;
 
   private boolean isPaused = false;
+  private boolean isCancelled = false;
```
dabeshouse: As we discussed, consider folding these into a status enum, but if it gets too convoluted, reconsider.

dabeshouse (commented):
A few comments. I think you had more changes planned?

dabeshouse (approved):
lgtm

PR #490, opened 2025-11-01
Remove usage of JScrollPaneCoveringLinuxPaintBug

dportnoy15:


dportnoy15:
I don't see any adverse effects in Linux with this code removed, although I haven't done a whole lot of testing yet. Also, do we want to remove other classes like the ones with PaintOmittingWhenAppropritate in the name?

core/util/src/main/java/edu/cmu/cs/dennisc/javax/swing/components/ScrollBarPaintOmittingWhenAppropriateJScrollPane.java
```diff
@@ -51,12 +53,24 @@
...
       super(orientation);
     }
 
+    @Override
+    public Color getBackground() {
+      javax.swing.JScrollPane scrollPane = (javax.swing.JScrollPane) this.getParent();
+      if (scrollPane != null) {
+        Color scrollPaneBackground = scrollPane.getBackground();
+        return scrollPaneBackground;
+        //todo?
```
dabeshouse: I see this is copied from JScrollPaneCoveringLinuxPaintBug.
Do you think this todo is worth preserving?
If it goes scrollPaneBackground can go too and this simplifies.

jennej:
i'm in favor of removing as many scrollpanes as possible

dabeshouse (commented):
Any idea what the original linux paint bug was?
How did you come across this corner of the code?

core/util/src/main/java/edu/cmu/cs/dennisc/javax/swing/components/ScrollBarPaintOmittingWhenAppropriateJScrollPane.java
jennej: I found this when I was scrolling through the dennisc.* files and deleting the unused ones

dportnoy15:
I started over, initially just removing some child classes of JScrollPaneCoveringLinuxPaintBug. Doing this, I do see a change in Linux, albeit a positive one. When those classes were used, such as in the class code view or gallery, the scroll bars and panes being scrolled would only update at certain positions, such as at 25% intervals. However, removing this code results in smooth scrolling and feels a lot nicer.

core/croquet/src/main/java/org/lgna/croquet/views/HorizontalScrollBarPaintOmittingWhenAppropriateScrollPane.java
```diff
@@ -42,7 +42,6 @@
  *******************************************************************************/
 package org.lgna.croquet.views;
 
-import edu.cmu.cs.dennisc.javax.swing.components.HorizontalScrollBarPaintOmittingWhenAppropriateJScrollPane;
 import edu.cmu.cs.dennisc.javax.swing.components.JScrollPaneCoveringLinuxPaintBug;
```
jennej: this import is unused, checkstyle will yell

dabeshouse (approved):
lgtm

PR #493, opened 2025-11-05
Bugfix/loop var

### LINKED ISSUE #491: Attempting to rename counter in for loop causes repeated exceptions
This occurs consistently using Alice 3.9.0.3 RC2 build 1285 on Windows 11. 

When using a "count"(for) loop in Java language mode, any attempt to rename the loop counter ("indexA") will result in an exception. This happens when rename is chosen after a right click on ANY of the occurrences of indexA in the loop definition. Once this exception is triggered, Alice becomes unstable and will continue to generate exceptions when nearly anything is clicked until a full restart of Alice.

dabeshouse:
Fix for java view loop variable renaming.
Fixes #491


core/story-api/src/main/java/org/lgna/project/ast/StaticAnalysisUtilities.java
```diff
@@ -55,24 +55,26 @@
...
+    }
+    char c0 = identifier.charAt(0);
+    if (!Character.isLetter(c0) && (c0 != '_')) {
+      // Must start with letter or underscore
+      return false;
+    }
+    for (int i = 1; i < N; i++) {
+      char cI = identifier.charAt(i);
+      if (!Character.isLetterOrDigit(cI) && (cI != '_')) {
+        // Must contain only letters, numbers, or underscores
```
jennej, line 73: comments and everything! 😍 

this is much clearer, thx

jennej (approved):
lgtm

PR #494, opened 2025-11-05
Drop event recorder that was written to but never read from.

dabeshouse:
We stored lots of events. We never look at them again. This stops doing that.

(Spotted while fixing loop var but putting as a separate PR.)


core/story-api/src/main/java/org/lgna/story/implementation/eventhandling/EventManager.java
```diff
@@ -149,18 +136,13 @@ public EventManager(SceneImp scene) {
...
     }
-    inputRecorder = new InputEventRecorder(scene);
     contingent = new TimerContingencyManager(timer);
   }
 
   public void initialize() {
     scene.addSceneActivationListener(timer);
   }
 
-  public MouseEventWrapper createWrapper(MouseEvent e) {
```
jennej, line 160: can we remove MouseEventWrapper too?

jennej (commented):
looks good, just looks like there is one more file we can delete

core/story-api/src/main/java/org/lgna/story/implementation/eventhandling/EventManager.java
dabeshouse: We can!

jennej (approved):

PR #498, opened 2025-11-08
Restore initializeInFrame

dabeshouse:
Upon reflection we can't remove code from SProgram that is run by reflection.


dportnoy15 (approved):

core/story-api/src/main/java/org/lgna/story/implementation/ProgramImp.java
```diff
@@ -345,6 +340,43 @@ public void initializeInAwtContainer(Container awtContainer) {
     this.initializeInAwtContainer(new DefaultAwtContainerInitializer(awtContainer));
   }
 
+  public void initializeInFrame(final JFrame frame, final Runnable runnable) {
```
jennej, line 343: can you add a comment about why these shouldn't be deleted in the future to maybe save future us?

jennej (changes requested):

jennej (approved):
thank you!

PR #499, opened 2025-11-10
simplifying the icon factory manager situation

jennej:
this is just some gentle moving and renaming

I think my next step is going to be to merge the StoryIconFactoryManager into the IconFactoryManager class, assuming we want to keep the FieldIcon* classes.


dabeshouse (approved):
lgtm

PR #502, opened 2025-11-11
Feature/gl crash

dabeshouse:
Fix OpenGL crashes.
Adjust some what thread some work is done on to better cleanup and limit log spew.
Some code cleaning.


jennej (approved):
lgtm

PR #503, opened 2025-11-11
the last of my icon cleanups

jennej:


core/ide/src/main/java/org/alice/stageide/StoryApiConfigurationManager.java
```diff
@@ -166,11 +164,7 @@ private static enum TypeComparator implements Comparator<AbstractType<?, ?, ?>>
 
     private double getValue(AbstractType<?, ?, ?> type) {
       Double value = mapTypeToValue.get(type);
-      if (value != null) {
-        return value;
-      } else {
-        return DEFAULT_VALUE;
-      }
+      return Objects.requireNonNullElse(value, DEFAULT_VALUE);
```
dabeshouse, line 167: Good to see this. We should do more.

core/ide/src/main/java/org/alice/stageide/StoryApiConfigurationManager.java
```diff
@@ -391,19 +365,12 @@ public boolean isTabClosable(AbstractCode code) {
...
-    //    rv.add( org.lgna.project.ast.JavaType.getInstance( org.lgna.story.Biped.class ) );
-    return rv;
-  }
-
   @Override
   protected List<? super JavaType> addSecondaryJavaTypes(List<? super JavaType> rv) {
     super.addSecondaryJavaTypes(rv);
 
     rv.add(JavaType.getInstance(SJoint.class));
+    // the nulls added to the list become line breaks in dialogs where you can choose a type
```
dabeshouse, line 373: We could define a constant line break and set it to null, if that makes anything clearer.

dabeshouse (approved):
Generally looks good.
So many nice deletions - and that dispatching code is inspired.

PR #505, opened 2025-11-18
laf cleanup tasks

jennej:
* slightly smarter icon setting
* fix how we display eventlisteners in dropdown menus so that it isn't light grey text on white

### DIFF
--- a/core/ide/src/main/java/org/alice/ide/croquet/models/cascade/ExpressionFillIn.java
+++ b/core/ide/src/main/java/org/alice/ide/croquet/models/cascade/ExpressionFillIn.java
@@ -92,7 +92,8 @@ public abstract class ExpressionFillIn<F extends Expression, B> extends Immutabl
     }
 
     JComponent expressionPane;
-    // to procedure block into our dropdown, make it look like a proper procedure block, but without the knurls
+    // to procedure block in our dropdowns, make it look like a proper procedure block, but without the knurls
+    // (Found in the 'Add Event Listener' dropdown on the initializeEventListeners tab)
     if (expression instanceof MethodInvocation invocation && invocation.method.getValue().isProcedure()) {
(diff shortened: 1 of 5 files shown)


core/ide/src/main/java/org/alice/ide/members/components/templates/ProcedureInvocationTemplate.java
```diff
@@ -65,13 +65,15 @@ protected void changed(ListPropertyEvent<UserParameter> e) {
...
-  /* package-private */ProcedureInvocationTemplate(AbstractMethod method) {
+  public ProcedureInvocationTemplate(AbstractMethod method) {
     super(ProcedureInvocationTemplateDragModel.getInstance(method));
     this.method = method;
 
     if (this.method instanceof UserMethod) {
       this.setPopupPrepModel(MethodTemplateMenuModel.getInstance((UserMethod) this.method).getPopupPrepModel());
     }
+
+    handleDisplayable();
```
dabeshouse, line 76: So you went with this instead of refresh.

core/ide/src/main/java/org/alice/ide/croquet/models/cascade/ExpressionFillIn.java
```diff
@@ -88,7 +90,21 @@ protected JComponent createMenuItemIconProxy(ItemNode<? super F, B> step) {
     } else {
       trailingLabel = null;
     }
-    JComponent expressionPane = PreviewAstI18nFactory.getInstance().createExpressionPane(expression).getAwtComponent();
+
+    JComponent expressionPane;
+    // to procedure block into our dropdown, make it look like a proper procedure block, but without the knurls
```
dabeshouse: Might be worth including a mention of the event listeners drop down, since that is what this is for.

dabeshouse (approved):
A suggested comment addition, otherwise lgtm.

PR #508, opened 2025-11-19
Seal Chunk and streamline switch

dabeshouse:
Code cleaning, or exercise, but I think it's worth keeping.


core/ide/src/main/java/org/alice/ide/i18n/MethodInvocationChunk.java
```diff
@@ -45,7 +45,7 @@
 /**
  * @author Dennis Cosgrove
  */
-public class MethodInvocationChunk extends Chunk {
+public final class MethodInvocationChunk extends Chunk {
   private String methodName;
```
dportnoy15: Why not make methodName final as well?

dportnoy15 (changes requested):

PR #511, opened 2025-11-21
Feature/glr tidy

dabeshouse:
The changed code is used when pick is not functioning correctly, so it mostly isn't.

The driver was GlrGeometry.getIntersectionInSourceFromPlaneInLocal that passed in x, y, x instead of x, y, z.


core/glrender/src/main/java/edu/cmu/cs/dennisc/render/gl/imp/adapters/GlrDisc.java
```diff
@@ -80,7 +81,7 @@ private void glDisc(Context c) {
 
   @Override
   public Point3 getIntersectionInSource(Ray ray, Matrix4x4 m, int subElement) {
```
jennej, line 83: this is certainly more clear

jennej (approved):

PR #515, opened 2025-12-15
Feature/menu justification

dabeshouse:
If the menu items all pretend to be max width they are left justified.
This is a hackish, I mean non-standard, way to do that.

There is also the beginnings of a project release mechanism and some code cleaning.

Targeted for develop, but might also go into release.

### DIFF
--- a/core/croquet/src/main/java/org/lgna/croquet/imp/cascade/RtItem.java
+++ b/core/croquet/src/main/java/org/lgna/croquet/imp/cascade/RtItem.java
@@ -220,10 +220,11 @@ abstract class RtItem<F, B, M extends CascadeItem<F, B>, C extends AbstractItemN
     }
   };
 
-  protected ViewController<?, ?> createMenuItem(CascadeItem<F, B> item, boolean isLast) {
+  protected ViewController<?, ?> createMenuItem() {
     ViewController<?, ?> rv;
     JMenuItem jMenuItem;
-    if (isLast) {
+    M item = getElement();
+    if (isLast()) {
       CascadeMenuItem menuItem = new CascadeMenuItem(item, this.getRtRoot());
       menuItem.getAwtComponent().addActionListener(this.actionListener);
       jMenuItem = menuItem.getAwtComponent();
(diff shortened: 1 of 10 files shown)


core/croquet/src/main/java/org/lgna/croquet/imp/cascade/RtItem.java
```diff
@@ -305,10 +305,10 @@ protected List<CascadeBlank<Void>> getModelBlanks() {
   }
 
   @Override
-  protected ViewController<?, ?> createMenuItem(CascadeItem<Void, Void> item, boolean isLast) {
-    //todo
+  protected ViewController<?, ?> createMenuItem() {
+    CascadeSeparator item =  getElement();
     if ((item.getMenuItemText() != null) || (item.getMenuItemIcon(null) != null)) {
-      ViewController<?, ?> rv = super.createMenuItem(item, isLast);
```
jennej, line 311: is isLast() on super going to be different?

core/ide/src/main/java/org/alice/ide/declarationseditor/TypeMenu.java
```diff
@@ -105,7 +107,12 @@ private TypeMenu(NamedUserType type) {
   @Override
   protected void localize() {
     super.localize();
-    this.setSmallIcon(new TypeIcon(this.type, true, TYPE_FONT, BONUS_FONT));
+    this.setSmallIcon(new TypeIcon(this.type, true, TYPE_FONT, BONUS_FONT) {
```
jennej, line 115: also add a comment about why you're doing this?

core/croquet/src/main/java/org/lgna/croquet/imp/cascade/RtItem.java
dabeshouse: A call to super calls the method on super, but the object remains the same.
If `RtSeparator.createMenu()` calls `super` on `RtItem` the `this` remains an `RtSeparator` and it will always return `true` for `isLast()`.

core/ide/src/main/java/org/alice/ide/declarationseditor/TypeMenu.java
dabeshouse: Yeah, this is a little odd and worth clarifying.

dportnoy15 (approved):

core/ide/src/main/java/org/alice/ide/declarationseditor/TypeMenu.java
dabeshouse: Comment added

PR #516, opened 2025-12-18
Fix preview rendering

dportnoy15:
Fix a bug where the world preview wouldn't resume rendering after running a world, if the world had been paused.

The order of setActiveScene() and stopAnimator() were originally changed to ensure everything was garbage collected properly after running a world in the editor, but this particular change no longer seems necessary for that.


dabeshouse (approved):
lgtm

PR #519, opened 2026-01-07
Match earlier scroll arrow display logic

dabeshouse:
Prior to 0751844c43684b87b0eda9949555545cec4318d2


jennej (approved):
comments and everything, it is beautiful

PR #521, opened 2026-01-12
Initialize rendering before passed in file is opened.

dabeshouse:
Move render initialization earlier to prevent locking up when opening a world on startup.


dabeshouse:
This helps for some worlds, but smaller worlds can still hit the race condition so I am withdrawing this and looking for a better fix.

PR #533, opened 2026-02-18
Laf/cleanup

jennej:
small fixes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -456,7 +456,7 @@
             <configuration>
               <rules>
               <requireMavenVersion>
-                <version>[3.9.9,4]</version>
+                <version>[3.9.9,4)</version>
                 <message>The build requires maven version 3.9.9.
   If you want to try another use:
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -456,7 +456,7 @@
             <configuration>
               <rules>
               <requireMavenVersion>
-                <version>[3.9.9]</version>
+                <version>[3.9.9,4]</version>
```
dabeshouse: `4]` implies it will build with 4. Unless you have verified that we should probably use `4)` for anything up to, but not including, 4.

dabeshouse (changes requested):
Commented code, improved rendering, all good.
One detail on maven version.

pom.xml
jennej: oh, i just copied and pasted without understanding. will )

dabeshouse (approved):
LGTM

PR #534, opened 2026-02-18
Feature/control palette

dabeshouse:
Visual tweaks for control flow palette.

### DIFF
--- a/alice-ide/src/main/resources/org/alice/stageide/themes/FlatLaf.properties
+++ b/alice-ide/src/main/resources/org/alice/stageide/themes/FlatLaf.properties
@@ -89,3 +89,6 @@ Alice.Block.knurlForeground = #D9D9D9
 # contrasting text/backgrounds
 Alice.Block.contrastForeground = #3E3E41
 Alice.Block.contrastBackground = #D9D9D9
+
+Alice.ControlFlow.knurlActive = #000000
+Alice.ControlFlow.knurlInactive = #333333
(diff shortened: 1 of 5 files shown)


dportnoy15 (approved):

core/ide/src/main/java/org/alice/ide/controlflow/components/MiniControlFlowStatementTemplate.java
```diff
@@ -146,4 +147,16 @@ protected void handleUndisplayable() {
     //    this.removeAllComponents();
     super.handleUndisplayable();
   }
+
```
jennej, line 150: is there a way to do this without the insets? i found unraveling the inset code really unintuitive.  also, why 6?

alice-ide/src/main/resources/org/alice/stageide/themes/FlatLaf.properties
```diff
@@ -88,4 +88,7 @@ Alice.Block.knurlForeground = #D9D9D9
 
 # contrasting text/backgrounds
 Alice.Block.contrastForeground = #3E3E41
-Alice.Block.contrastBackground = #D9D9D9
\ No newline at end of file
+Alice.Block.contrastBackground = #D9D9D9
+
+Alice.ControlFlow.knurlActive = #000000
```
jennej, line 93: i don't love this, but I need more time to think about how I'd do it differently.

I would expect this to be under Alice.Block tho

Issue #535, opened 2026-02-20
Animation error on Windows for certain worlds


dabeshouse:
The attached world triggers the error on 3.9.0.3.

[platform.a3p.zip](https://github.com/user-attachments/files/25451049/platform.a3p.zip)

```
java.lang.NullPointerException: Cannot invoke "org.lgna.story.implementation.ProgramImp.getAnimator()" because the return value of "org.lgna.story.implementation.SceneImp.getProgram()" is null
at org.lgna.story.implementation.eventhandling.TimerEventHandler$1.automaticDisplayCompleted([TimerEventHandler.java:72](http://timereventhandler.java:72/))
at [edu.cmu.cs.dennisc.render.gl](http://edu.cmu.cs.dennisc.render.gl/).GlrRenderFactory.handleDisplayed([GlrRenderFactory.java:216](http://glrrenderfactory.java:216/))
at [edu.cmu.cs.dennisc.render.gl](http://edu.cmu.cs.dennisc.render.gl/).GlrRenderFactory.step([GlrRenderFactory.java:151](http://glrrenderfactory.java:151/))
at [edu.cmu.cs.dennisc.render.gl](http://edu.cmu.cs.dennisc.render.gl/).GlrRenderFactory$1.step([GlrRenderFactory.java:176](http://glrrenderfactory.java:176/))
at [edu.cmu.cs.dennisc.render.gl.Animator.run](http://edu.cmu.cs.dennisc.render.gl.animator.run/)([Animator.java:121](http://animator.java:121/))
at java.base/java.lang.Thread.run([Thread.java:1583](http://thread.java:1583/))
```

PR #536, opened 2026-02-23
Fix an animation error on some worlds (Github Issue #535)

dportnoy15:


dabeshouse (approved):
lgtm
As we discussed, I will post a belt and suspenders PR with project null checks

PR #537, opened 2026-02-24
Fix/null program

dabeshouse:
Belt and suspenders fixes for null program in sceneImp. Plus some code cleanup

### DIFF
--- a/core/ide/src/main/java/org/lgna/ik/poser/scene/PoserPicturePlaneInteraction.java
+++ b/core/ide/src/main/java/org/lgna/ik/poser/scene/PoserPicturePlaneInteraction.java
@@ -61,6 +61,7 @@ import org.lgna.story.SMovableTurnable;
 import org.lgna.story.SSphere;
 import org.lgna.story.implementation.CameraImp;
 import org.lgna.story.implementation.EntityImp;
+import org.lgna.story.implementation.ProgramImp;
 import org.lgna.story.implementation.SceneImp;
 
 import javax.swing.SwingUtilities;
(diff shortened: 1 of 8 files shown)


core/ide/src/main/java/org/lgna/ik/poser/scene/PoserPicturePlaneInteraction.java
```diff
@@ -244,14 +245,16 @@ private JointSelectionSphere pickJoint(JointSelectionSphere one, JointSelectionS
...
-    SceneImp implementation = scene.getImplementation();
-    RenderTarget rt = implementation.getProgram().getOnscreenRenderTarget();
-    PickResult pickResult = rt.getSynchronousPicker().pickFrontMost(e.getPoint(), PickSubElementPolicy.NOT_REQUIRED);
-    if ((pickResult != null) && (pickResult.getVisual() != null)) {
-      Composite composite = pickResult.getVisual().getParent();
-      if (composite != null) {
-        if (composite.getParent() instanceof ManipulationHandle3D) {
-          return (ManipulationHandle3D) composite.getParent();
+    ProgramImp program = scene.getImplementation().getProgram();
+    if (program != null) {
```
dportnoy15: Perhaps instead do this instead, if you're ok with early returns, so the main part of the code doesn't need to be indented:

```
if (program == null) {
   return;
}
```

core/story-api/src/main/java/org/lgna/story/implementation/eventhandling/EventManager.java
```diff
@@ -239,12 +238,14 @@ public void sceneActivated() {
...
     if (this.dragAdapter == null) {
       this.dragAdapter = new RuntimeDragAdapter(visuals);
-      OnscreenRenderTarget renderTarget = this.scene.getProgram().getOnscreenRenderTarget();
-      SymmetricPerspectiveCamera camera = (SymmetricPerspectiveCamera) scene.findFirstCamera().getSgCamera();
-      this.dragAdapter.setOnscreenRenderTarget(renderTarget);
-      this.dragAdapter.addCameraView(CameraView.MAIN, camera);
-      this.dragAdapter.makeCameraActive(camera);
-      this.dragAdapter.setAnimator(this.scene.getProgram().getAnimator());
+      ProgramImp program = scene.getProgram();
+      if (program != null) {
```
dportnoy15: Ditto about early returns

core/story-api/src/main/java/org/lgna/story/implementation/SceneImp.java
```diff
@@ -70,7 +70,7 @@
  * @author Dennis Cosgrove
  */
 public class SceneImp extends EntityImp {
-  private static final Transformable createDirectionalLightTransformable(DirectionalLight sgDirectionalLight, Angle yaw, Angle pitch, float brightness) {
+  private static Transformable createDirectionalLightTransformable(DirectionalLight sgDirectionalLight, Angle yaw, Angle pitch, float brightness) {
```
dportnoy15, line 73: Was the removal of final here intentional? Just wondering since I don't see the method being overridden anywhere in the PR.

core/story-api/src/main/java/org/lgna/story/implementation/SceneImp.java
```diff
@@ -192,41 +214,6 @@ public ProgramImp getProgram() {
...
-      this.program = program;
-      if (program != null) {
-        this.eventManager.addListenersTo(program.getOnscreenRenderTarget());
-      }
-    }
-    this.program = program;
-  }
-
-  //todo
-  //  private static class Capsule {
```
dportnoy15, line 209: Oh, no! My favorite class is going away!

dportnoy15 (approved):
Looks good! Just some minor suggestions about doing early returns.

core/story-api/src/main/java/org/lgna/story/implementation/SceneImp.java
dabeshouse: Yes. It is static, so final adds nothing.

core/ide/src/main/java/org/lgna/ik/poser/scene/PoserPicturePlaneInteraction.java
dabeshouse: Done

core/story-api/src/main/java/org/lgna/story/implementation/eventhandling/EventManager.java
dabeshouse: Done

PR #538, opened 2026-02-24
Feature/mini controls

dabeshouse:
Reworked enough I am making a new PR

### DIFF
--- a/core/croquet/src/main/java/org/lgna/croquet/views/MutableList.java
+++ b/core/croquet/src/main/java/org/lgna/croquet/views/MutableList.java
@@ -153,7 +153,7 @@ public abstract class MutableList<E> extends SwingComponentView<JPanel> {
       g2.drawRoundRect(0, 0, width, height, 8, 8);
 
       g2.setStroke(prevStroke);
-       g2.setPaint(KNURL_COLOR);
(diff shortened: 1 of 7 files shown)


core/croquet/src/main/java/org/lgna/croquet/views/MutableList.java
```diff
@@ -153,7 +153,7 @@ protected void paintComponent(Graphics g) {
       g2.drawRoundRect(0, 0, width, height, 8, 8);
 
       g2.setStroke(prevStroke);
```
jennej, line 155: this could be ThemeUtilities.getActiveTheme().getKnurlColorFor(BASE_COLOR);

jennej (changes requested):
just the one tweak, but this looks good

core/croquet/src/main/java/org/lgna/croquet/views/MutableList.java
dabeshouse: Except that ThemeUtilities is in the `ide` module and this is `croquet` which `ide` depends on.

dabeshouse: We discussed this in Slack so I am going call it good enough and merge

PR #543, opened 2026-03-16
Handle inability to load backup files in a more flexible way

dportnoy15:


dportnoy15:
This is meant to handle issues when various directories related to backups don't have write permissions, specifically the foolowing:

 - MyProjects not having write permissions
 - MyProjects/.defaultbak not having write permissions
 - Loading a project outside of MyProjects, when the parent folder of that project doesn't have write permissions

This is a WIP. So far, I just tested on Linux to make sure it handles those cases better.

core/ide/src/main/java/org/alice/ide/ProjectApplication.java
```diff
@@ -510,7 +515,12 @@ private void handleProjectLoadSuccess(Project project, File projectFile, UserAct
     } else if (unloadableFiles.isEmpty() && !uriProjectLoader.isNewProject()) {
       // check for backups newer than the project
 
-      File backupDir = projectFileUtilities.appropriateBackupDirectory(projectFile).toFile();
+      Path backupPath = projectFileUtilities.appropriateBackupDirectory(projectFile);
+      if (backupPath == null) {
+        return;
```
dabeshouse: Returning here halts the project loading.
The user has been notified there will not be backups, but they should still be able to load the project.

dabeshouse (changes requested):
We should be able to read files that cannot be backed up. This change just gives up.
Take a look at the conditional above as well. That empty if block could go.

dabeshouse (changes requested):
This is better. Now the user can load even if the backup directory is not writable.
They are warned about this when they load.

However they are then warned again every time Alice tries to make a backup, even stacking up dialogs if they are not dismissed.
If we can't write we should stop trying to.

dabeshouse (approved):

PR #550, opened 2026-04-08
Event listener ui fix

jennej:


dportnoy15 (approved):

core/ide/src/main/java/org/alice/ide/x/components/KeyedArgumentView.java
```diff
@@ -59,4 +66,25 @@ protected String getName() {
...
+  @Override
+  protected void internalRefresh() {
+    super.internalRefresh();
+    setBorder(new FlatRoundBorder());
+
+    // the code that sets the names of arguments is the same code that sets the names on our draggable
+    // procedures and functions (their color is set to Alice.Block.foreground). This won't work for our events
+    // (which afaict is the only thing that creates this particular view) so this code digs out the appropriate
+    // JLabel so that we can make it a readable color.
+    if (this.getAwtComponent() != null
```
dabeshouse: This can't be the way we have to do it. Can it?
I am trying to find another way. When I give up I will return and merge this in.

dabeshouse: I have given up. Merging.

